PR #10, opened 2015-01-21
Add UIKit import to generated file

nmccann:
After following the installation instructions I encountered build errors informing me that `UIImage` and `UIStoryboard` were unknown types. I've update the code generator to include an import of `UIKit` to resolve these errors.

Now, I would imagine this issue would have come up earlier if it's affecting everyone - so maybe I missed something that would have resolved this without code changes.

Also, I thought I should inform you that the idea behind this project is genius - I thought that I would always have to deal with storyboards/images as strings, I didn't know there was a better way until now, and I love the better way! Thank you!

### DIFF
--- a/R.swift/func.swift
+++ b/R.swift/func.swift
@@ -139,6 +139,10 @@ func writeResourceFile(code: String, toFolderURL folderURL: NSURL) {
 
 // MARK: Code generator functions
 
+func swiftImports() -> String {
+    return "import UIKit"
+}
+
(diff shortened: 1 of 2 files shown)


mac-cain13:
Thanks for contributing and the great compliment @nmccann!

I think it depends on if `UIKit` is already linked somewhere else in the project and that is hit earlier by the compiler. But surely we shouldn't depend on that, so I'll merge this in!

PR #23, opened 2015-02-17
Fixed R.validate line breaks

ReneLindhorst:


mac-cain13:
Thanks for using R.swift and taking the time to create a PR for this issue!

I've decided to use a slightly different solution (see: https://github.com/mac-cain13/R.swift/commit/ea48b1adec042dfa15da19d6500dbcc34af8fe8a) I think that's a bit more elegant.

Once again thanks for contributing in this way, much appreciated!

PR #24, opened 2015-02-17
Changed access control to public

ReneLindhorst:
so that e.g. unit tests can use the generated code


ReneLindhorst:
Well, this might not be the best idea in case of internal types.

Issue #21, opened 2015-02-02
Typed overload of dequeueReusableCellWithIdentifier


mac-cain13:
Would be nice to write this:
`let x = tableView.dequeueReusableCellWithIdentifier(R.reuseIdentifier.rerouteBusinessHoursCell, forIndexPath: indexPath)`

instead of this:
`let x = tableView.dequeueReusableCellWithIdentifier(R.reuseIdentifier.rerouteBusinessHoursCell, forIndexPath: indexPath) as RerouteBusinessHoursCell`

tomlokhorst:
I'm not sure you can do that. However you could definitely write something like:

```swift
let x = tableView.dequeueReuesableCellRerouteBusinessHours(forIndexPath: indexPath)
```

Although I'm not sure I like "polluting" UITableView with a bunch of extension methods.
Maybe put the functions on the generated code, something like:

```swift
let x = R.reuseIdentifier.dequeueRerouteBusinessHourCell(withTableView: tableView, forIndexPath: indexPath)
```

tomlokhorst:
Now that I think about it, you can of course do what you originally wrote, using a phantom type:

```swift
struct ReuseIdentifier<T> : Printable {
  let value: String

  var description: String {
    return value
  }
}

extension UITableView {
  func dequeueReusableCellWithIdentifier<T : UITableViewCell>(identifier: ReuseIdentifier<T>, forIndexPath indexPath: NSIndexPath) -> T? {
    return dequeueReusableCellWithIdentifier(identifier.value, forIndexPath: indexPath) as? T
  }
}
```

And by having R.swift generate:

```swift
struct R {
  struct reuseIdentifier {
    static var rerouteBusinessHoursCell: ReuseIdentifier<RerouteBusinessHoursCell> { return ReuseIdentifier(value: "RerouteBusinessHoursCell") }
  }
}
```

mac-cain13:
Closing this, moving discussion to the pull request

PR #28, opened 2015-03-25
Better reuse identifiers

mac-cain13:
Attempt to implement issue #21

Current concerns:
- [x] Changing reuseIdentifiers to objects makes it hard to use them as strings, maybe we should make a seperate category of this "higher level" R.swift feature?
(I want it to be simple to start using R.swift and not force users to use our higher level functions as this could mean being not able to use R values in certain edge cases.)

- [x] CollectionView support missing
- [x] TableView header/footer support missing
- [x] Registering nibs should also be less messy

### DIFF
--- a/R.swift/func.swift
+++ b/R.swift/func.swift
@@ -150,29 +150,65 @@ func nibVarForNib(nib: Nib) -> Var {
 }
 
 func nibStructForNib(nib: Nib) -> Struct {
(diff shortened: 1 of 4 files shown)


mac-cain13:
Changing reuseIdentifiers doesn't seem to be that hard since we now have overloads for those functions. :)

PR #44, opened 2015-08-04
Feature/tests

mac-cain13:
*This pull request:*
- Adds integration tests by adding a example project with tests that runs R.swift and checks if stuff from the generated file works. First basics tests are there.
- Adds unit tests by adding tests to the R.swift binary target, the 3 first basic tests are there
- Converts R.swift itself to Swift 2(!)
- Converts R.swift generated file to Swift 2(!)
- Fixes an issue where the AppIcon was also generated as an R.image resource, but it will always return nil
- Adds warnings for images with duplicate names

*Todo:*
- [x] Bitrise should run the tests automatically
- [ ] Write more tests (PR can be merged before this is finished)


mac-cain13:
Just realized Travis will not work with XCode 7 for a while. :(
Maybe look into other options or just wait out, will sleep a night over that. :)

petester42:
You  should check out [bitrise](https://www.bitrise.io) as a travis alternative. They keep up with the latest Xcode versions.

mac-cain13:
@petester42 Thanks for suggesting Bitrise, we're using it for some other internal projects at Q42. I should really try to set it up for this project too I think indeed. Thanks again! :)

PR #46, opened 2015-08-05
Add support for Swift 2 (and Carthage)

AndrewSB:
I naively added a dynamic framework as a build target for R.swift to make it compatible with [Carthage](https://github.com/Carthage/Carthage). After debugging for a few minutes I realized that R.swift isn't actually used at runtime at all - it's just an executable used at build time - so there's no way to support Carthage 😢

I left the work I did to add Carthage compatibility in this project, if you'd like I can remove those commits and leave in just the Swift 2 ones (fork is 100% compatible with Swift 2, but I'd appreciate a quick lookover)

Absolutely :heart:LOVE:heart: the project. Turning (imo) the last weakly typed piece of swift into strong typed goodness.
Thank you so much @mac-cain13


mac-cain13:
Thanks for all the praise and the time you put into this! Good to know you have checked the Carthage integration option. R.swift is not a framework so sounds plausible it's not suited for this project. :)

I also now have PR https://github.com/mac-cain13/R.swift/pull/44 with tests and also the switch to Swift 2. Looks like we more or less did the same work, sadly. :(

Will merge that PR later and close this one so we also have some tests!

AndrewSB:
Was thinking about this some more - Can't use Carthage for the rswift executable that's used as a build phase to find resources, but could definitely use it within the code when you reference something of the sort `R.storyboard.[storyboardname]`...

What do you think?  
Would still need to setup a hook in Carthage to run and build the executable on clone, but I'm sure the community would love to be able to use R.swift as a dynamic framework :smile:

AndrewSB:
w.r.t. Swift conversion, should have forked off #44 - my mistake

mac-cain13:
No worries about the forking, #44 was only available moments before you submitted your PR. :)

About Carthage, do you feel there is any real advantage of using Carthage in that scenario? I don't really want to use Carthage for the sake of using Carthage. There should be a real benefit in it, maybe I'm not seeing something here you do see?

AndrewSB:
Carthage allows you to pull in a library as a dynamic framework, which is preferred by some. You get some cool features like not loading the library into address space until it's actually needed, namespacing, etc. 
[Carthage talks a little about it's philosophy](https://github.com/Carthage/Carthage#differences-between-carthage-and-cocoapods), and I personally like the decentralized, non integrated way Carthage allows me to manage my dependencies.

R.Swift could definitely be imported as a dynamic framework (through Carthage) for the `R.storyboard.[storyboardname]` use case in swift code, I'm still not sure how the executable used on build would be handled, since that's a product that needs to be built by `xcodebuild`. 
If you'd like I can look further into that, there may be a way to move the executable into the framework/ force the executable to be built by Carthage.

Also, importing a library that doesn't have a shared dynamic framework makes it significantly less fun to use Carthage, I'd love to help out to get it integrated 🙌

AndrewSB:
Actually - Disregard what I said above. Took a closer look at the project and realized that the way it works is through the struct in the `.generated.swift` file. 
I'll try to see if it's possible to hook into Carthage's `build` to create the executable and let you know.

Nevertheless, if it doesn't make sense to merge this into the project, can it be maintained (I volunteer :stuck_out_tongue_closed_eyes:) on a branch? I'd prefer to not include it in my project from my fork and I think other people who'd like to use Carthage would appreciate the option

mac-cain13:
Thanks for taking the time for those posts! :)

Long story short; I'd love to integrate Carthage if it provides something useful for Carthage users. So please go ahead and create a PR if you think it can be useful! I'll happy to take a look at it, discuss it and merge it in if it's a nice addition for the majority of the Carthage users!

If it won't make it in you can always maintain it as a seperate project. :)

AndrewSB:
I'll create a new PR :) thanks for being understanding @mac-cain13

AndrewSB:
So after a serious look at how Carthage works, I don't really think it's possible to build the executable without manually opening the project and `⌘B`ing.

I still think it would be awesome to either keep the framework target in the project or maintain a `carthage` branch. It'll make it so people who want to use the library with R.swift don't have to maintain a fork.

Let me know what you think @mac-cain13, I'd be happy to maintain the branch if you'd like and depending on the decision I can add a Carthage installation guide to the README

mac-cain13:
I would say create the branch and give me a ping over here. I'll take a look and then we can decide what the best option for it is; a branch, a fork or something else.

daniel-shin1104:
+1 for Carthage Support. Frankly I've had enough with all the under-the-hood complexities (and bugs) that cocoapods bring in. Cocoapods simply does too much as a **dependency manager** unlike other great ones like pip, npm, gem etc. Carthage just does that. No magic. Everything that it does is crystal-clear and thus much easier to work with when you have dozens of libraries pulled in.

daniel-shin1104:
And great work on R. This is just what I needed coming from Android :tada:

PR #47, opened 2015-08-05
modify if statement for Swift 1.2 syntax

minsOne:
Hello,  Swift 1.2 is supported `where` on if statement.  then I changed some if statement.

Thanks


mac-cain13:
Good improvement thanks! Will definitely merge this in, but I'm also in the process of converting to Swift 2 in PR https://github.com/mac-cain13/R.swift/pull/44, so will combine it. :)

mac-cain13:
Merged your commit into PR #44, when that one is merged into master your commits will also be taken into the new release. :)

Thanks again!

PR #52, opened 2015-08-07
Support nibs with spaces in name

tomekc:
Hello there!
I have a nib in my project named "Launch Screen.xib", which I am pretty sure has been generated by Xcode. 

R.swift generated uncompilable code, because using file name as a symbol. Here's a quick fix to differentiate nib name and variable name. Hope it is fine for you :)

### DIFF
--- a/R.swift/func.swift
+++ b/R.swift/func.swift
@@ -149,8 +149,8 @@ func internalNibStructFromNibs(nibs: [Nib]) -> Struct {
 }
 
 func nibVarForNib(nib: Nib) -> Var {
-  let structType = Type(name: "_R.nib._\(nib.name)")
-  return Var(isStatic: true, name: nib.name, type: structType, getter: "return \(structType)()")
+  let structType = Type(name: "_R.nib._\(nib.symbolName)")
+  return Var(isStatic: true, name: nib.symbolName, type: structType, getter: "return \(structType)()")
 }
 
 func nibStructForNib(nib: Nib) -> Struct {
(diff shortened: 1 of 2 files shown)


mac-cain13:
Thanks for using R.swift and taking the time to fix this issue!

This sounds like something that is fixed in a quite recent release. The `Var` struct is using the `sanitizedSwiftName` function and in recent releases this will also strip out spaces of names. So that should already fix this issue.

Could you check which version of R.swift your using and if the current 0.8.4 release has this already fixed? Would be good to know!

Once again thanks for your time!

tomekc:
Hi! I used cocoa pods and believe I used the most recent version. 

Interesting, that I forked from master (as you may check) and it didn't work either - which eventually made me creating this patch.

tomekc:
I think I got it. Fix you mentioned was correcting the variable name, not a struct type. 

Actual code generated was:
```
static var launchScreen: _R.nib._Launch Screen { return _R.nib._Launch Screen() }
```

I probably should have used `sanitizedSwiftName` function instead of making my own.

mac-cain13:
Aah yeah I see! Thanks for clearing this up! :)

Indeed, the `sanitizedSwiftName` function should be used and I also want to write a test for this one. Will do that on develop and the next release will fix this. Thanks again for reporting.

mac-cain13:
I've created a seperate issue for this one, will solve it over there and close this PR!

mac-cain13:
Fixed this in 0.8.5!

Issue #38, opened 2015-07-28
Type fonts


ntnmrndn:
On iOS, user can provide custom fonts through Info.plist's UIAppFonts.
It would be nice to provide completion for:

     func R.font.appFonts.fontName(fontSize:14.0) -> UIFont!
     var R.font.appFonts.fontName: String

Bonus:

    R.font.system.helvetica(fontSize:14.0) -> UIFont!
    var R.font.system.helvetica: String

This is a bit complicated since with have to find the right names for the fonts, but it would be a really nice feature :)

Maybe I can find some time to implement that.
Can you give me a feedback before I start ?

mac-cain13:
Agreed, that would be awesome! Great idea. :)

I would prefer `R.font.myOwnFont(fontSize: 14.0) -> UIFont?`. This is a bit shorter and R.swift never force unwraps anything, the library should not crash if the resource is not available.

System fonts could also be under `R.font.anIosFontName(fontSize:14.0) -> UIFont?` since duplicate font names will not work anyway. :)

If you have time to create a PR please go ahead! Keep me posted about your state. :)

ntnmrndn:
Good point, `R.font` would be enough.

mac-cain13:
Looked a bit into this and I think that the `fc-query` commanline tool can help us out to get the fontname that we have to use in iOS.

`fc-query --format "%{family}" [fontfile]` will report a list of fontnames in this file. So far the last one in the list seemed to be the correct one the use. Look like it is `fontname` or `familyname,fontname`, but it seems that all kind of crazy combinations are possible, so it could be that we're wrong.

ntnmrndn:
Good job !
fc-query is definitely the way to go, and it should be trivial to integrate the fonts now.
        `find -E .  -regex ".*\.(ttf|otf)" -exec fc-query {} -f "%{postscriptname}\n" \;`
Using this command display the correct names of all fonts in sub directories.
I'll do some PR by the end of next week if you don't do it before :)

ntnmrndn:
I wrote a draft which is returning font names as Strings.
I used `fc-query` at `"/opt/X11/bin/fc-query"` path.

Since `fc-query` seems to be an optional binary, I'm a bit puzzled about wether it is safe to use a hardcoded path.
Reading PATHS from environment might be a better solution, but I'm wondering wether this is the way to go or not.

We could also fail and ignore font generation if fc-query is not at the expected path (my preferred solution).

[...]

PR #66, opened 2015-08-12
Add support for custom fonts

ntnmrndn:
From https://github.com/mac-cain13/R.swift/issues/38
Finally, no need to use options or weird path.
Does not yet support system fonts (needs quite some extra work to add available annotations to R.swift).


mac-cain13:
@ntnmrndn Question; Why do you give a default value for the size of `0`? The docs state that you should give a size *greater* than zero. So that doesn't seem like a good default to me?!

ntnmrndn:
It was a mistake. I had the belief that "fontSize 0" had the special meaning to keep current fontSize, when assigning an UIFont, but I can't find this anymore.

PR #70, opened 2015-08-20
Carthage support for Swift 2

AndrewSB:
Finally got it done :smile:

### DIFF
--- a/R.swift.xcodeproj/project.pbxproj
+++ b/R.swift.xcodeproj/project.pbxproj
@@ -7,6 +7,11 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


mac-cain13:
Thanks for being involved and creating this PR!

I've looked into your PR and I see two changes:
- Creation of a Framework target just to make Carthage not complain, it's not really used.
- Instructions to install with Carthage, where the user basically uses Carthage only to download the source and then has to compile the binary themselves.

I'm sorry, but I only see downsides with this approach;
- It's more complex then just manually installation
- Carthage will build the framework and people will be confused what they should so with is
- The unused Framework makes the project harder to understand for contributors and users

Therefore I won't include this Carthage PR into the project. As the people of the Carthage project stated; Carthage is for building dynamic frameworks, not to integrate binary tools like R.swift.

Carthage is not the right tool for the job for this project, right now. So we won't support Carthage for now.

Issue #71, opened 2015-08-25
Move Segue identifiers into ViewController extensions and make them typed


mac-cain13:
See https://github.com/mac-cain13/R.swift/issues/34#issuecomment-134694637:

> Segues should be changed so they are an extension on the viewcontroller and not global things. This makes duplicates less likely and if in that situation there are duplicates they really are a problem so a warning is appropriate then.
> 
> This also would enable us to make typed segues in the end.

tomlokhorst:
All other strongly typed resources are under the global `R.*`, no?

Would this change this to something along the lines of `myViewController.R.segue.openDetail`?
Or simply introduce an extra level of namespace: `R.segue.MyViewController.openDetail`.

mac-cain13:
Yeah, have to give this some thought about how it exactly should work, but a segue is really something that is linked to the source view controller. But I think something like:

```
class SomeViewController {
  func someMethod() {
    performSegue(Rsegues.toSettingsSegue) { typedDestinationController in
       something()
    }
  }
}```

Would be great, but as I said have to think about it. The exception here is that nearly all resources are "global" and segues aren't.... What would you prefer @tomlokhorst?

tomlokhorst:
Is the only reason for adding segues to ViewControllers so that they can be strongly typed to the destination? It's a good reason to be sure, but otherwise there is no need to, correct?

Note that to do the performSegue with a callback, you need to also have code in the `prepareForSegue` method. I think it would always require work from the user, it can't all be code generated. See [SegueManager](https://github.com/tomlokhorst/SegueManager/).

mac-cain13:
Typed segues are well underway now, let's close this one and move all discussion to https://github.com/mac-cain13/R.swift/pull/80

PR #74, opened 2015-09-03
Work on typed segues

tomlokhorst:
_This pull request is a work in progress._

I've started the work for typed segues as discussed in https://github.com/mac-cain13/R.swift/issues/71
This isn't complete yet (XML Storyboard parsing isn't done).

Question: Do we want to remove the existing `R.segue.something` segues in favour of the `R.segue.MyViewController.something`?
It seems weird to duplicate them, although the newer typed ones are more complicated and this would be a backward compatibility breaking change.

Todo:

 - [ ] Adapt Storyboard XML parser to generate `Storyboard.Segue` values
 - [ ] Create test cases

### DIFF
--- a/R.swift/func.swift
+++ b/R.swift/func.swift
@@ -276,7 +276,7 @@ func varFromReusable(reusable: Reusable) -> Var {
   return Var(
     isStatic: true,
     name: reusable.identifier,
-    type: ReuseIdentifier.type.withGenericType(reusable.type),
+    type: ReuseIdentifier.type.withGenericArgs([reusable.type.name]),
     getter: "return \(ReuseIdentifier.type.name)(identifier: \"\(reusable.identifier)\")"
   )
 }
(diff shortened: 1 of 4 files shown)


mac-cain13:
I think it's okay to break in this case. 0.9.0 will be a breaking upgrade anyway as I want to parse the xcodeproj file instead of scanning folders in this version.

What do you think about making an extension on the specific view controller so you can type: `MyViewController.segues.something` for example?

tomlokhorst:
All other uses of R.swift go through the global `R.*` namespace, I don't see a benefit in changing that here.

Why do you want to use an extension on the specific controller classes?

Note that this could even potentially break if there's already a `segues` property on `MyViewController`.

tomlokhorst:
(Unrelated to this specific PR)
Can Bitrise.io be set to publicly display its output?
Now I see on Github "All checks have failed", but I cannot see why.

mac-cain13:
Hmm not sure on Bitrise, it currently fails because they haven't upgraded to Xcode beta 6 yet and our Swift 2 code is not compatible with beta 5. :)

mac-cain13:
Let's continue in PR #80 on this one. :)

PR #80, opened 2015-09-13
Feature/typed segues

mac-cain13:
_Note:_ This PR supersedes PR #74.

I'm still in doubt if I shouldn't make an extension on the VC since `performSegue(segues.toForgotPassword)` seems so much nicer and less error prone to me than `performSegue(R.segues.ThisViewController.toForgotPassword)`. Especially the `.ThisViewController` is something I don't like.

Maybe it should exist both and then you should be able to turn of the extension in case of a conflict for example.

**Todo:**
- [x] Handle view controllers that are in multiple storyboards with segues that may/may not have the same identifier/destination by warning about them as duplicate and not generating any code for them
- [x] Check what happens with a segue to a modally presented `UINavigationController`, should we type the root view controller in there? Is it logical the way it works right now? -> These can't be given an identifier, so those are out of scope
- [x] Create overload of `performSegue` for ease of use without special stuff
- [x] Create overload/method that actually use type information
- [x] Cleanup code
- [ ] Create tests
- [x] Test storyboard references -> Works for now, all are `UIViewController` now, when picking up issue #120 this will be fixed
- [x] Update documentation


mac-cain13:
@tomlokhorst What is you idea about VCs that are in multiple storyboards with different or overlapping segues? Should we add an extra namespace layer? Like `R.segue.mainStoryboard.SomeViewController.ThisSegue` or do you see another solution?

Also; What is your opinion on what R.swift should provide to make use of this new typed segues?

tomlokhorst:
Re: multiple storyboards with overlapping segues.
What is the current behaviour when you do `self.performSegue("segueName")`?
If it has undefined behaviour or an error, I think R.swift should just generate a warning en no segue, similarly to other duplicate identifiers.

For usage, R.swift should generate something like:

```
func isSegue(segue: UIStoryboardSegue) -> Bool {
  return segue.identifier == self.identifier
}

func destinationViewController(segue: UIStoryboardSegue) -> Destination? {
  return isSegue(segue) ? segue.destinationViewController as? Destination : nil
}

**Comment by tomlokhorst** [comment]
As part of this;

If the segue is actually configured to be a custom subclass of UIStoryboardSegue, perhaps R.swift should also generate something of that specific subclass?

I haven't thought about this a lot, so maybe this doesn't make sense in this context.

**Comment by mac-cain13** [comment]
Only two more checkboxes to tick!

File: R.swift/values.swift [hunk 1]
```diff
@@ -42,6 +42,75 @@ let ReuseIdentifier = Struct(
...
+  ],
+  functions: [
+    Initializer(
+      type: .Designated,
+      isFailable: true,
+      parameters: [
+        Function.Parameter(name: "segue", type: Type._UIStoryboardSegue)
+      ],
+      body: [
+        "guard let segue = segue as? Segue, sourceViewController = segue.sourceViewController as? Source, destinationViewController = segue.destinationViewController as? Destination else { return nil }",
```
tomlokhorst: Shouldn't this also check the `identifier`?

mac-cain13: Good point. Can't do the check here, but will at it in the `typeSegue()` method.

mac-cain13:
- [ ] Rewrite `R.segue.myVC.mySegue.typeSegue(segue)` to `segue.typedForIdentifier(R.segue.myVC.mySegue)`

mac-cain13:
I feel like this is good enough now to merge into develop and give it a try on some projects.

PR #101, opened 2015-10-09
The productName in the pbxproj doesn't seem to get updated. Matching with name instead. Requires Xcode.swift update.

kylejm:
I noticed this today while building another target in the same project
as the original target I’ve been using R.swift with:

If I run `rswift`, I get an error saying that there is not a target
with the name matching the target I’m running, then it prints the
targets name:

```
[R.swift] Target ‘MyAppStaging’ not found in project file, available
targets are: MyApp, MyAppTests, MyApp  // This second 'MyApp' is MyAppStaging
```

Which is obviously not the case because I’m running said target.

I took a look into the `.pbxproj` and found that `My App Staging`’s
`productName` was still set to the `productName` that I had duplicated
it from (My App).

I checked Xcode build settings several times but couldn’t see why this
was the case. It seems that Xcode doesn’t really use this value and
only sets it when creating the project? (I have the build settings
product name set for this Staging target, but it’s not reflected here
by `productName`…).

It is, however, reflected by `name`, so targets should be matched
against name instead.

This commit requires an update on the Xcode.swift module:

https://github.com/mac-cain13/Xcode.swift/pull/1

### DIFF
--- a/R.swift/types.swift
+++ b/R.swift/types.swift
@@ -339,7 +339,7 @@ struct Xcodeproj {
   func resourceURLsForTarget(targetName: String, pathResolver: Path -> NSURL) throws -> [NSURL] {
     // Look for target in project file
     let allTargets = projectFile.project.targets
-    guard let target = allTargets.filter({ $0.productName == targetName }).first else {
+    guard let target = allTargets.filter({ $0.name == targetName }).first else {
       let availableTargets = allTargets.map { $0.productName }.joinWithSeparator(", ")
       throw ResourceParsingError.ParsingFailed("Target '\(targetName)' not found in project file, available targets are: \(availableTargets)")
     }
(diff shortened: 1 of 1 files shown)


tomlokhorst:
Thanks for looking into this and fixing the problem! (I'm the creator of Xcode.swift)

I think the next line that generates the error message should also be updated to use the new `name` instead of `productName`:

```swift
let availableTargets = allTargets.map { $0.name }.joinWithSeparator(", ")
```

kylejm:
No problem :blush: 

Good spot, I'll for that now!

mac-cain13:
Finally merged your PR into `develop`!

Will be merged into master with the next release. :)

kylejm:
Awesome! Thank you!

Does this mean that I won't show up as a contributor?

mac-cain13:
You will show up as a contributor as soon as I merge it into master. That will be when I release 0.10.1 with your fix in it. :)

No worries! :)

kylejm:
Awesome! Thank you!! :blush:

Issue #105, opened 2015-10-28
tvOS support


tomlokhorst:
Similar to [Mac OS X support](https://github.com/mac-cain13/R.swift/issues/102), it would be nice to have support for tvOS.

Issue #102, opened 2015-10-12
macOS support


ghost:
Now that OS X has storyboards, having statically typed storyboards and segues would be quite useful, not even mentioning added type-safety for images and other assets

PR #106, opened 2015-10-28
Support tvOS in podspec

tomlokhorst:
I think this is all that's needed to add tvOS support to R.swift (https://github.com/mac-cain13/R.swift/issues/105).

I also looked into adding Mac OSX support (https://github.com/mac-cain13/R.swift/issues/102), but that's a bit more work.
R.swift is currently hardcoded to using UIKit, that would need to be made dynamic to also support AppKit.


mac-cain13:
Awesome!

PR #108, opened 2015-11-01
Release 0.11

mac-cain13:
- Correct @2x/3x loading for non-PNG files and device specific suffixes like ~ipad and ~iphone
- Use correct product name during build (Thanks @kylejm)
- Support tvOS in podspec (Thanks @tomlokhorst for testing basic tvOS support)
- Better cleanup of invalid characters in filenames when converting to variable names, also preserve capitals for better readability


kylejm:
\o/

PR #110, opened 2015-11-04
Feature/tabbar relation

tomasharkema:
I've created a typesafe way to get a certain ViewController from a UITabBarController.

Works as follows:

```swift
someTabBarInstance.viewController(R.storyboard.main.mainTabBarController.firstViewControllerRelation)
```

or

```swift
let controllerFromTabBar = R.storyboard.main.mainTabBarController.firstViewController
```


R.swift/func.swift
```diff
@@ -151,6 +152,48 @@ func storyboardStructForStoryboard(storyboard: Storyboard) -> Struct {
...
+          return Var(isStatic: true, name: relatedViewController.storyboardIdentifier ?? "\(ordinal)ViewContoller", type: relatedViewController.type.asOptional(), getter: "return instance!.viewControllers![\(idx)]\(getterCast)")
+        }
+
+        let relationsVars = zip(relatedViewControllers, Ordinals).map { (related, ordinal) -> Var in
+          let (idx, relatedViewController) = related
+          return Var(
+            isStatic: true,
+            name: (relatedViewController.storyboardIdentifier ?? "\(ordinal)ViewContoller") + "Relation",
+            type: Type(name: "TabRelation", genericType: relatedViewController.type, optional: false),
+            getter: "return TabRelation<\(relatedViewController.type.description)>(identifier: \"asdfa\", index: \(idx))")
```
mac-cain13: What's this `identifier: \"asdfa\"`?

mac-cain13:
Thanks for using R.swift and contributing!

I think we should take similar approach as with segues in PR https://github.com/mac-cain13/R.swift/pull/80, here you now do: 
```
func prepareForSegue(segue: UIStoryboardSegue, sender: AnyObject) {
  if let segue = R.segue.myViewController.someSegue.typeSegue(segue) {
    let typedDestinationVC = segue.destinationViewController
  }
}
```

We could do this for the tabbar, we create a `R.storyboard.main.mainTabBarController.typeTabBar(tabBarController: UITabBarcontroller)` method. You call that to get a `TypedTabBarController` type back that contains typed properties such as `firstTabController` (up to `fifthTabController`) and possibly other controllers.

I think this is:
- More in line with how segues now do it, creating a more uniform character
- The method would fail if one tab can't be converted, gives more assurance everything is okay at once
- It feels clearer to me than the currently proposed solution.

mac-cain13:
I'll close this PR since I'm going to delete the `develop` branch in favor of shorter running branches. Also working towards 1.0 I'm going to do a major restructure of the project that will break this PR.

I think this is a good first shot at typing tabbars and I'll like a "cast helper" thingy for tabbars in R.swift. So I'll create an issue instead of this PR and give that another look later on.

Thanks again for your effort and contributing to R.swift!

PR #114, opened 2015-11-28
Use availability check and iOS7 fallback for UIImage

iv-mexx:
`UIImage.init(named:inBundle:compatibleWithTraitCollection:)` is only available in iOS 8.0 and later.

By using swifts new `#available(iOS 8.0, *)` check and `UIImage(named:<name>)` as fallback, R.swift can also be used for on iOS7. 

There's one caveat though: `UIImage(named:<name>)` can only load images from the main bundle. I think that this is OK (and it's documented in the Q&A) because this is mostly relevant when using dynamic frameworks (e.g. in 3rd party swift dependencies) which is not supported on iOS7 anyway.

### DIFF
--- a/Documentation/QandA.md
+++ b/Documentation/QandA.md
@@ -23,3 +23,8 @@ During installation you add R.swift as a Build phase to your target, basically t
 - Every time you build R.swift will be runned
 - It takes a look at your Xcode project file and inspects all resources linked with the target currently build
 - It generates a `R.generated.swift` file that contains a struct with types references to all resources
+
+## Does R.swift work for iOS7?
+
+R.swift does also work for iOS7, with one caveat though: 
+- Images can only be loaded from the main bundle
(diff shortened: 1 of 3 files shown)


mac-cain13:
Thanks for contributing! I'll merge this in. 

Note that I don't really like exceptions like these that make your R.swift work on iOS 7, but not in all cases. If it become too many I'll drop iOS 7 support all together, but for now I think this is a nice fix to keep iOS 7 on board.

iv-mexx:
> Note that I don't really like exceptions like these that make your R.swift work on iOS 7, but not in all cases. If it become too many I'll drop iOS 7 support all together, but for now I think this is a nice fix to keep iOS 7 on board.

I understand, and at first I was hesitant myself. I tried it in a fairly large project that still needs to support iOS7 and this was the only issue that came up. Of course, that does not mean that everything is covered though.

mac-cain13:
Just realize that this will generate a warning for each image you have when your deployment target is iOS 8 or higher.

I see two solutions:
1. Drop iOS 7 support.
2. Check the deployment target and generate different code.

Maybe I'll implement solution 2 for now to get 0.13 out of the door and then drop iOS 7 support when we split the types out to a different library.

Personally I think that Swift with iOS 7 support is quite a hassle anyway and with the number I'm seeing for our apps we should consider dropping it.

Issue #122, opened 2015-12-03
Move R.swift types to library


tomlokhorst:
R.swift currently has some types that are used in the generated code in `R.generated.swift`.

For example: `ReuseIdentifier`, `NibResource`, and the recently introduced `TypedStoryboardSegueInfo`.

It would be nice if these are moved to a R.swift library, instead of being in the generated file. That way other libraries can extend R.swift by using these types and creating extensions.

mac-cain13:
If you want to discuss this one please feel free to do so in #123.

PR #123, opened 2015-12-04
Move R.swift types to library project

mac-cain13:
This implements #122.

Goals:
- Make it easy for other libs to extend our types
- Integration with Cocoapods should stay as easy as it is now
- Clean up R.swift code

Thing to tackle:
- iOS 7 compatibility
- Integration without Cocoapods
- ...


mac-cain13:
*Integration of library with users project*
I'm considering to put the library in a seperate repo. This would make it easier for Carthage, Cocoapod and people without a package manager to add the library to their project.

In the current situation Cocoapods will work just fine by making the library a new pod and depending on it, but Carthage can't build it because of all the schemes and projects that are in this repo. Without a projectmanager the seperate repo could contain instructions on how to include this.

*iOS 7 compatibility / using R.swift without the lib*
First idea was to create a flag that will inline the types into the `R.generated.swift` file. This would not make you able to use extensions based on the library, but will enable you to use R.swift on iOS 7. For iOS 7 this is an okay solution I think. You can't use frameworks anyway.

For iOS 8+ it is quite a downside and it would clutter installation instructions and options. So I don't really feel users will benefit from this option building for those targets. I think when this PR merges iOS 7 support should be dropped and we _shouldn't_ build the flag.

The library will then be mandatory to use, and instructions on how to include it can be added.

tomlokhorst:
I suggest removing the `UIStoryboardSegue` and `UIViewController` restrictions from `StoryboardSegueIdentifier` as well as the `UIViewController` restrictions on `TypedStoryboardSegueInfo`.

These restrictions aren't used in the types and add a lot of overhead, for example when creating extensions. External users of those type (liked the generated `_R` type), can still impose these restrictions themselves.

mac-cain13:
Good point @tomlokhorst. This would also make it compatible with OS X in the future.

tomlokhorst:
Not exactly, since `TypedStoryboardSegueInfo` actually uses a `UIStoryboardSegue` in its initializer.
Although, maybe that can be moved to an extension.

mac-cain13:
Yeah, I was just looking at that, I think all UIKit stuff should be in a extension file. That way we can subspec it etc.

Issue #111, opened 2015-11-05
Add UICollectionViewCell reuse example in the docs


sxua:
It would be nice to add a reuse identifiers for UICollectionView cells.

mac-cain13:
It sure would, was one of the initial reasons I started the project. I think this should be explicitly stated in the docs to prevent confusion if it's in there.

Thanks for using R.swift!

PR #139, opened 2015-12-27
Add collection view cell example

tomlokhorst:
Implements https://github.com/mac-cain13/R.swift/issues/111


mac-cain13:
:+1: Thanks

PR #159, opened 2016-02-05
Synthesizing let accessors for storyboard identifiers

JaviSoto:


mac-cain13:
Thanks for the PR! I'll take a look as soon as I have a bit of time to review. :)

JaviSoto:
:+1:

R.swift/SwiftTypes/Property.swift
```diff
@@ -12,6 +12,26 @@ protocol Property: UsedTypesProvider, CustomStringConvertible {
   var name: String { get }
 }
 
+struct AnyProperty: Property {
```
JaviSoto: I implemented type-erasure here so that the compiler doesn't complain when putting `Let`s in `[Property]` arrays

mac-cain13:
Looked at the PR. Thanks again for creating it!

I have a few questions;
- What is your use case that you need to retrieve the identifier of a VC instead of the VC itself?
- You're using the type name instead of the identifier name for the generated property, this could conflict if I have 2 VCs of the same type in 1 storyboard, both with an identifier. Shouldn't we use the identifier so it becomes:

```
let avpVC = "avpVC"
func avpVC(_: Void) -> AVPlayerViewController? {
        return UIStoryboard(resource: self).instantiateViewControllerWithIdentifier(self. avpVC) as? AVPlayerViewController
}
```

R.swift/SwiftTypes/Property.swift
mac-cain13: The type-erasure is a nice solution, I would just use a function here I think:

```
func anyProperty(property: Property) -> Property {
  return property
}
```

Bit simpler and easier to maintain, but not a big deal. Will change this myself when the PR is merged and apply the type-erasure also in other parts of the code.

JaviSoto: I'll change this!

JaviSoto:
Sorry for the late reply!

>What is your use case that you need to retrieve the identifier of a VC instead of the VC itself?

I have a custom set up so that I don't instantiate the VCs from UIStoryboard, but instead each VC has its own initializer to do proper compile-time-safe dependency injection. For this I need the identifier (to pass it to a common parent class that instantiates the controller, which is written in Obj-C cause Swift doesn't allow to return anything other than self from an init)

>You're using the type name instead of the identifier name for the generated property, this could conflict if I have 2 VCs of the same type in 1 storyboard, both with an identifier

I remove duplicates if that's the case. Is it normal to have 2 of the same type in one storyboard? I can change it to the identifier though. I remember I did that initially and changed it, but I can't remember why.

JaviSoto:
>I remove duplicates if that's the case. Is it normal to have 2 of the same type in one storyboard? I can change it to the identifier though. I remember I did that initially and changed it, but I can't remember why.

I made that change :) let me know how it looks!

mac-cain13:
Thanks for the update! Little busy over here, but I'll take a better look asap and see if I can merge it in! :)

Issue #60, opened 2015-08-09
Homebrew


larcus94:
Would be cool if we could install this using Homebrew :)
Then we wouldn't need to have the binary in every single project folder.

mac-cain13:
Thanks for your suggestion!

Yeah would be handy for devs working on their own on smaller projects, but devs in bigger teams with multiple projects will run in a sort of DLL-hell like problem where you have version 0.9 of R.swift for project A that is in Swift 2 and I have version 0.7 for project B that I can't upgrade because that needs to be Swift1.2 for a while.

Putting the R.swift binary (that I want to convert to a Swift script btw instead of a binary, see #36) in the project folder makes sure every devs is using the correct version for that project. It also guarantees you can downgrade just one project in the case that R.swift generates invalid code for your project.

So I feel it's better to see these kind of tools as "libraries" that you compose your project of and not as a tool that is installed on you Mac globally. Let me know if you disagree and have good arguments for the homebrew approach that I'm missing!

larcus94:
Hmm well that doesn't really contradict your use case. You can still have it in the projects folder if you like, which makes sense for big teams. For smaller teams however, it's probably handier to install it on the computer.
With Homebrew it's really simple to downgrade to a specific version as well if something went wrong. It essentially works just like Cocoapods.

mac-cain13:
True, it would be possible to downgrade, but then again all project are downgraded/upgraded at once.

Good point that you could still use R.swift in a project if you prefer that method. That would be absolutely true.

I don't really see the big benefit when using Homebrew. Since new developers would have to install one more tool before getting their project up and running and taking the risk of a versioning problem.

What would be your main argument for using Homebrew instead of embedding in your project?

larcus94:
It's not a huge benefit, admittedly. It's just more comfortable and more what people are used to. R.swift would be the first tool I'd install in my project folder, opposed to Carthage, Cocoapods and mogenerator. 
So it wouldn't make a huge difference but would allow people to use and manage their tools the way they are used to.

larcus94:
Another advantage is that as a formula, it's easier to update R.swift itself. Then I can just update it together with any other package.

uny:
+1

AndrewSB:
+1 would love to see on homebrew

mac-cain13:
Okay, I'm convinced there are users that would use R.swift like this. I'll look into creating a R.swift formula for R.swift. :)

AndrewSB:
[...]

PR #166, opened 2016-02-11
Add homebrew documentation to readme

tomasharkema:
as of https://github.com/mac-cain13/R.swift/issues/60, r.swift is available in the homebrew package manager.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -82,12 +82,31 @@ _Tip:_ Add the `*.generated.swift` pattern to your `.gitignore` file to prevent 
 ### Manually
 
 0. Add the [R.swift.Library](https://github.com/mac-cain13/R.swift.Library#Installation) to your project
-1. [Download](https://github.com/mac-cain13/R.swift/releases) a R.swift release, unzip it and put it into your source root directory
+1. [Download](https://github.com/mac-cain13/R.swift/releases) a R.swift release, unzip it and put it into your source root directory OR install the binary via homebrew. For more detail check [Homebrew](#Homebrew)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -82,13 +82,32 @@ _Tip:_ Add the `*.generated.swift` pattern to your `.gitignore` file to prevent
 ### Manually
 
 0. Add the [R.swift.Library](https://github.com/mac-cain13/R.swift.Library#Installation) to your project
-1. [Download](https://github.com/mac-cain13/R.swift/releases) a R.swift release, unzip it and put it into your source root directory
+1. [Download](https://github.com/mac-cain13/R.swift/releases) a R.swift release, unzip it and put it into your source root directory OR install the binary via homebrew. For more detail check [Homebrew](#homebrew)
```
mac-cain13: The installation instruction in step 3 is using `"$SRCROOT/rswift"`, that won't work when you use Homebrew.

mac-cain13:
Thanks for the PR and making R.swift available on Homebrew! :)

The manual installation edit will not work using homebrew, because of the path, I would suggest to make the `### Homebrew` section just like `Cocoapods` and `Manual` and put a step by step instruction. The path the the `rswift` binary is different.

Also the whole output of the `rswift -h` call is something I would take out, just saying that you can call `rswift -h` is enough I think. (We'll definitely will forget to updated the readme when the help output changes...)

tomasharkema:
Maybe some input for a discussion: should we replace the manual installment via homebrew? or

mac-cain13:
Ping @tomasharkema.

Are you going to give this PR an update?

Issue #137, opened 2015-12-21
Sanitizing names for use in swift should warn on empty string


mac-cain13:
Sanitizing names for use in swift should warn on empty string, for example if you have a reuse identifier ` ` or `:)` then this will generate invalid code. R.swift should warn about this.

Issue #136, opened 2015-12-16
Refactor warnings so that it's not an sideeffect anymore


mac-cain13:
Functions should just return warnings they have encountered. Later on we can print them at once.

This makes code clearer, easier to test and gives us the ability to suppress warnings. We could even add meta data so specific types of warnings could be shown/hidden etc.

tomlokhorst:
Plan of attack for part of this, related to identifier warnings:

 - [ ] Create a separate `SwiftIdentifier` type instead of a string
 - [ ] Refactor all data types (Struct, Let, etc) to use `SwiftIdentifier`
 - [ ] Refactor code in SanitizedSwiftName.swift to use `SwiftIdentifier`
 - [ ] Move all code related to identifier checking (duplicates, empty names) into a single (parameterised function), instead of repeating it across all validators. This should fix https://github.com/mac-cain13/R.swift/issues/175

Continuing on:

 - [ ] Identify a list of all other warning types
 - [ ] Create an enum for those
 - [ ] Figure out how to return those best from functions (return value, `inout` param, or mutable object passed around?)

PR #172, opened 2016-03-11
Don't generate code for empty identifier

tomlokhorst:
Implements https://github.com/mac-cain13/R.swift/issues/137

No longer generate (invalid) code for sanitized Swift identifiers that are an empty string. Instead print a warning. 

The code for printing warnings is a bit repetitive, and can be cleaned up. This falls under https://github.com/mac-cain13/R.swift/issues/136, I think.

### DIFF
--- a/R.swift/Generators/ImageGenerator.swift
+++ b/R.swift/Generators/ImageGenerator.swift
@@ -63,8 +63,7 @@ struct ImageGenerator: Generator {
     let groupedFunctions = allFunctions.groupBySwiftNames { $0.callName }
 
     for (sanitizedName, duplicates) in groupedFunctions.duplicates {
-      let names = duplicates.map { $0.name }.sort().joinWithSeparator(", ")
-      warn("Skipping \(duplicates.count) images because symbol '\(sanitizedName)' would be generated for all of these images: \(names)")
+      warn("Skipping \(duplicates.count) images because symbol '\(sanitizedName)' would be generated for all of these images: \(duplicates.joinWithSeparator(", "))")
     }
 
     let imageLets = groupedFunctions
--- a/R.swift/Generators/ImageGenerator.swift
+++ b/R.swift/Generators/ImageGenerator.swift
@@ -59,15 +59,16 @@ struct ImageGenerator: Generator {
         )
       }
 
-    let functions = (assetFolderImageFunctions + imageFunctions)
-      .groupUniquesAndDuplicates { $0.callName }
+    let allFunctions = assetFolderImageFunctions + imageFunctions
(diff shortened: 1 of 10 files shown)


R.swift/Generators/ImageGenerator.swift
```diff
@@ -59,15 +59,23 @@ struct ImageGenerator: Generator {
         )
       }
 
-    let functions = (assetFolderImageFunctions + imageFunctions)
-      .groupUniquesAndDuplicates { $0.callName }
+    let allFunctions = assetFolderImageFunctions + imageFunctions
+    let groupedFunctions = allFunctions.groupBySwiftNames { $0.callName }
```
tomlokhorst: Both `.callName` and `.groupBySwiftNames` call `sanitizeSwiftName`, so this name is sanitized twice. This should not be a problem, I think.

R.swift/Generators/SegueGenerator.swift
```diff
@@ -41,19 +41,31 @@ struct SegueGenerator: Generator {
...
-      .groupBy { $0.sourceType }
-      .values
-      .flatMap(SegueGenerator.seguesWithInfoForSourceTypeToStruct)
+      for (name, duplicates) in gropuedSeguesWithInfo.duplicates {
+        warn("Skipping \(duplicates.count) segues for '\(sourceType)' because symbol '\(name)' would be generated for all of these segues, but with a different destination or segue type: \(duplicates.joinWithSeparator(", "))")
+      }
+
+      let empties = gropuedSeguesWithInfo.empties
+      if let empty = empties.first where empties.count == 1 {
+        warn("Skipping 1 segue for '\(sourceType)' because no swift identifier can be generated for segue: \(empty)")
```
tomlokhorst: The "empty" warning is printed once per sourceType. Might be nicer to print it for all segues at the same time.

R.swift/Generators/ImageGenerator.swift
mac-cain13: `sanitizeSwiftName` should be idempotent so that shouldn't give any problems. :)

R.swift/Generators/SegueGenerator.swift
mac-cain13: Would be a nice future improvement I think. It wouldn't be a warning that you get a lot of at once.

mac-cain13:
Nice PR, thanks! :+1: 
I'll merge it in.

I'm also considering a `SwiftName` type that `Function`, `Var` and `Let` etc take for their `name` property. That makes it clearer whether we are using an escaped value or not. Could be an upgrade of this PR.

What do you think?

Issue #125, opened 2015-12-05
See if and how Swiftdoc comments can help in autocompletion


mac-cain13:

mac-cain13:
Just tried it, it's quite nice to document stuff, really helps in autocompletion and inline help. Good thing to do wherever we can.

mac-cain13:
First version is merged, we can now expand where handy. :)

PR #173, opened 2016-03-11
Generate comments

tomlokhorst:
Implements https://github.com/mac-cain13/R.swift/issues/125

Generate comments for some of the generated code.
Not everything is done yet, this is a work in progress.


tomlokhorst:
A future refinement can be to generate references to images inline in the comments, like so:

screen shot 2016-03-12 at 00 18 34

Issue #169, opened 2016-02-22
Add support for R.colors from .clr files


tomlokhorst:
Mentioned by @renrawnalon in https://github.com/mac-cain13/R.swift/issues/41#issuecomment-187021929 R.swift could support generating colors from .clr files.

See also: https://www.natashatherobot.com/xcode-color-palette/

Although the exact workflow for adding a color pallete to R.colors isn't clear.
Since the Xcode project itself doesn't reference the .clr file.
And R.swift can't just use everything from `~/Library/Colors/*.clr`, because the R.generated.swift file would then be different for each member of a team.

Maybe the user should pass a directory of .clr files to the `rswift` executable from the Build Phase? That seems less than ideal.

renrawnalon:
Thanks for creating this issue. I've also considered how colors with fit in with the R.swift ecosystem, and for the reason you mentioned where the Xcode project doesn't reference .clr files, it doesn't seem to fit in. This also relates to one problem that I'm having with using .clr files, which is that it's difficult to put the .clr file for a project under version control. There are a few reasons for this:

1. .clr files are binary files, so it's not possible to git to merge changes from multiple people. As far as I know, the only way to parse these binary files is with a mac app using the NSColorList class as described [here](http://stackoverflow.com/questions/11386513/how-to-create-osx-clr-palette-files).
2. When a change is made to a color palette via Xcode, the ~/Library/Colors/Project.clr file is completely overridden. This breaks any symlinks to the file, so that isn't a reasonable solution.

Our current solution is to keep a copy of the file checked into out project, then manually copy it to and from ~/Library/Colors when changes are made. Every time this is done, a restart of Xcode is required to refresh the color palette in IB. This is obviously not an ideal solution.

Since this problem is not really in the scope of the problem that R.swift attempts to solve, I don't know if R.colors should be included in the main R.swift library. Does the idea of creating a sub-library for R.color like "R.swift.color" that takes the path to a color.clr file for directory somewhere in a project as a parameter sound like overkill?

tomlokhorst:
Thinking about this some more. I think most of the problems here are outside the scope of R.swift to solve.

One way to solve the "multiple people edit the same color palette"-problem; Use a tool that maps a .clr-file to a textual representation (two-way). That way, the text file can be checked in, and any conflicts can easily be resolved.
For the "installation in user directory"-problem; A build step could copy the .clr files from the repository to ~/Library/Colors and show a warning.

[...]

PR #174, opened 2016-03-13
Parse color palette files

tomlokhorst:
Implements https://github.com/mac-cain13/R.swift/issues/169

- [x] Add comments to generated code (including hex string)
- [x] Add tests to check if all conversions between colors works correctly


mac-cain13:
Looking good to me! :+1: 

One question; What about duplicates? Aswel filenames that result in duplicate identifiers as color names within the files that could create duplicate identifiers.

That's the only thing that maybe could go wrong I think.

tomlokhorst:
Whoops, forgot about those, I've added the code to warn about duplicate identifiers.

dxshindeo:
So is this working? Just added R.swift to my project, but it generates 0 colors (I have no named colors, but I do have custom color palette). Do I have to somehow specify this custom color palette so R.swift generates colors from it?

dxshindeo:
And the reason it was 0 was because: https://github.com/mac-cain13/R.swift/issues/214

tomlokhorst:
Note that `R.clr` is deprecated in favour of `R.color`:
https://github.com/mac-cain13/R.swift/blob/master/Documentation/Migration.md#upgrading-to-40

Support for color palettes will be removed in a future version of R.swift.

dxshindeo:
_tries not to cry
cries alot_

Thanks for info!

Issue #41, opened 2015-08-03
Support localized strings like R.strings.myString


Ewg777:
Add new feature parsing .strings file

mac-cain13:
Thanks for this interesting suggestion! Could you explain a little bit more about your vision on how this would work?

I think lot of developers type something like `NSLocalizedString("Please login when you're ready!", comment: "Login screen header")` and then extract that to a `.strings` file. That's not a case where R.swift would be really handy I think.

When using keys in localized strings maybe it's handier? A concrete example of how you would like it to work would be nice!

Ewg777:
For example like this http://developer.android.com/guide/topics/resources/string-resource.html

mac-cain13:
Yeah, I get you're inspired by the Android way. I believe we should think a little more about how this works out and integrates into the daily workflow of developers, but I definitly like the idea!

Ewg777:
It's already implemented https://github.com/AliSoftware/SwiftGen#localizablestrings

mac-cain13:
Ah nice example! Thanks for posting it here! :)

Definitely something to look into and implement in R.swift.

mac-cain13:
After talking with @tomlokhorst about this; I think that it will be hard to get a good solution that fits within the workflow. For example the SwiftGen style of working with string doesn't really shot how they add new strings.

zxy198717:
@mac-cain13 Could you generate string code like this:
```
struct string {
    static var bar_title: String? { return CustomLocalizedString("bar_title") } 
}
```
Then, create one file to implement function `CustomLocalizedString`, and developer can modify this file.
```
public func CustomLocalizedString(key: String) -> String {
    // DEV coding ....
    return NSLocalizedString(key, comment: "");
}
```

Ewg777:
[JiriTrecak/Laurine](http://github.com/JiriTrecak/Laurine) also does this trick

ntnmrndn:
Let's not forget strings dictionaries if we do that :)

dlbuckley:
I'm all for localised string generation, but I think there should be a way to access the string key separately so we can still use NSLocalizedString(). It's very useful to keep the macro/method as Xcode has tools that generate files based on it and it allows you to add more context to what you are using the text for.

I haven't tested it yet, but Xcode 7.3 also has string localisation testing in the static analyser, so thats another reason to keep the string keys available for reference.

tomlokhorst:
[...]

PR #176, opened 2016-03-15
[WIP] Parse Localizable.strings files

renrawnalon:
Implements #41 

- [x] Support single Localizable.strings file used as a string constants file.
- [x] Support multiple Localizable.strings files in their respective .lproj directories.
- [ ] Support printf style string formatting.
- [ ] Support .stringsdict files.


R.swift/Generators/StringsGenerator.swift
```diff
@@ -0,0 +1,51 @@
...
+      implements: [],
+      typealiasses: [],
+      properties: [],
+      functions: [StringsGenerator.localizedStringFunction()],
+      structs: []
+    )
+  }
+
+  private static func localizedStringFunction() -> Function {
+    return Function(
```
tomlokhorst: What is the goal of this function?

Why not use `NSLocalizedString` directly in the generated R.strings struct?

tomlokhorst:
This is really cool!

One extra feature we could consider is to copy the comment from the .strings file into the output file:

Something like:
```
/* Insert Element menu item */
"insert_element" = "Insert An Element";
/* Error string used for unknown error types. */
"ErrorString_1" = "An unknown error occurred.";
```

R.generated.swift:
```swift
  struct string {
    /// Insert Element menu item
    /// English: "Insert An Element"
    static let insert_element = NSLocalizedString("insert_element", comment: "Insert Element menu item")

    /// Error string used for unknown error types.
    /// English: "An unknown error occurred."
    static let errorString_1 = NSLocalizedString("ErrorString_1", comment: "Error string used for unknown error types.")
```

R.swift/Generators/StringsGenerator.swift
renrawnalon: This is preparation for when printf style formatting is supported and a list of arguments can be taken. I suppose adding it now is a bit premature, and even in the case where arguments are passed in, all of them could just be put on one long line. I thought separating that logic out would make the `R.generated.swift` file more readable.

tomlokhorst: This is a bit of a brother issue, I've created a general issue here: https://github.com/mac-cain13/R.swift/issues/177

renrawnalon:
Thanks for the feedback! I really like the comment idea. I was thinking about which language to display in the comment, and came to the conclusion that it would probably best to show the developer language. Apparently, this info, along with a list of all of the localized regions in the app, is provided in the xcodeproj in the following format:

```
developmentRegion = English;
hasScannedForEncodings = 0;
knownRegions = (
    en,
    Base,
    es,
    ja,
);
```

I think accessing this info and using to to generate comments shouldn't be too hard, but it will take me a little bit of time. Do this sound reasonable?

renrawnalon:
I'm not sure, but it looks like this build is failing because the Podfile.lock is out of date and ColorResource.swift isn't being added to project after updating the R.swift.Library submodule. After running `pod install`, the resource app builds and tests pass. Is this updated to Podfile.lock something that I should commit in this PR?

mac-cain13:
First of all, many thanks for taking a shot at implementing localized strings @renrawnalon! Awesome to see someone take the time to dive into this. I still have to take a better look on the code you wrote, I will asap! 

On generating a comment; Great idea, really like it. Adding an example string in the comment is also nice, but that could also be a nice upgrade in a future PR. I wouldn't spend to much time into that. I'd rather support stringsdict for example.

On the failing build; Seems that `ResourceGenerator.swift` won't compile because of `Use of unresolved identifier 'StringsGenerator'`. I think something else broke, you should be able to reproduce this by running the unit tests for the R.swift binary locally.

Feel free to push commit the updated lockfile if it resolves the issue. I can always fix it if needed during a merge to master. So whatever works. ;)

renrawnalon:
Thanks! I'll take a look!

tomlokhorst:
@renrawnalon I've added the fields `developmentRegion`, `hasScannedForEncodings`, and `knownRegions` to the develop branch of [Xcode.swift](https://github.com/tomlokhorst/Xcode.swift).

If you update the R.swift git submodule these three fields are available in `PBXProject`.

renrawnalon:
Great! I'll take a look and start working on the code to produce comments.

renrawnalon:
I'm not sure why this is failing on the ci environment. After I

1. Freshly clone this branch.
2. Run `git submodule update --init` 
3. Run `pod install`

everything builds fine and tests all execute without error on my local environment. Am I missing something? Can anyone else reproduce this build failure?

tomlokhorst:
The `StringsGenerator.swift` file isn't part of the test target, that's why it won't build.

Add `StringsGenerator.swift` and `Strings.swift` to `rswiftTests` by checking this second box for both files:

screen shot 2016-03-22 at 06 55 12

renrawnalon:
Sorry, I completely missed that @mac-cain13 told me how to reproduce the problem.

> you should be able to reproduce this by running the unit tests for the R.swift binary locally.

Everything seems to be working now. Thanks.

mac-cain13:
@renrawnalon How do you feel about this PR? It has 2 todo items open, is that something you are going to implement and we need before merging it in or is it in your opinion good to go as is?

If you consider it ready te be merged in and released for everybody I'll take some time to merge it, if you'd like to fix those other 2 points first I'll wait for that.

renrawnalon:
Sorry for the late reply. I've been busy these past couple of weeks. 

I think that it will take me a little time to get the 2 todo items finished, but the current implementation can't offer the same functionality that libraries like SwiftGen and Laurine, so it might be better to wait for the merge until these 2 items are finished. I'll go ahead and change this pull request to WIP to avoid confusion.

mac-cain13:
Let's close this one in favor of the PR @tomlokhorst continued to work on.

PR #181, opened 2016-03-24
Fix warnings in Xcode 7.3

chillpop:
Issues fixed:
- The workspace reference to the R.swift.Library project was an absolute path instead of a relative path.
- The ResourceAppTests target still had leftover Framework search paths from Xcode 6
- Swift 2.2 deprecated ++/-- operators
- Swift 2.2 deprecated function currying syntax

Let me know if you want me to bump the podspec version.

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - R.swift.Library (1.1.0)
+  - R.swift.Library (2.0.0)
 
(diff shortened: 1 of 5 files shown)


mac-cain13:
Thanks for the fixes! 

No need to bump the podspec/release a new version this is only code internal to the project. So users won't see these warnings, they will ship with the next release. :)

Issue #182, opened 2016-03-25
Add support for loading Apple TV parallax images


chillpop:
- Add support for parsing "imagestack" folders from asset libraries
 - Ignore "imageset" folders within "imagestacklayer" folders to avoid generating duplicate symbol warnings for images named 'Content' (eg: "Skipping 9 images because symbol 'content' would be generated for all of these images: Content, Content, Content, Content, Content, Content, Content, Content, Content")
- Add support for "lcr" and "lsr" parallax image stacks in app bundles

This would require changes to AssetFolder.swift and Image.swift.  It would probably be best to add a tvOS target to the ResourceApp example project for testing.

chillpop:
Note: I'm happy to make these changes but wanted to create an issue to make my intentions known

PR #183, opened 2016-03-27
Support for loading Apple TV parallax image assets

chillpop:
This addresses #182 

- "imagestack" folders within asset libraries are now included as image assets.
- "imagestacklayer" and "brandassets" folders are ignored to avoid generating warnings on tvOS projects.
- added a new Apple TV target (ResourceApp-tvOS) to test loading "imagestack" assets.

I tried to load .lsr and .lcr images included in the app bundle outside an asset catalog but that apparently doesn't work (or isn't supported).


mac-cain13:
Sounds amazing! Thanks for the PR, I'll take a closer look asap!

mac-cain13:
I've merged this in, still have to create a release with this new feature in it. Thanks again for taking the time to create a PR! Will create a release anytime soon. :)

Issue #175, opened 2016-03-14
Validation methods don't take duplicate identifiers into account


mac-cain13:
As @tomlokhorst mentioned to me.

tomlokhorst:
This could be fixed as part of https://github.com/mac-cain13/R.swift/issues/136

mac-cain13:
This is fixed now!

Issue #155, opened 2016-01-20
Add support for localised storyboards, nibs, and other resources


ikovalisko:
At the moment generator skips resources with the same file name. But this limits us with localizing storyboards and nibs. Would be great if R.swift support them. What are the caveats?

mac-cain13:
Very good point, what we need to do this:
- Detect when something is a different version of a localized nib (not sure if the projectfile makes that clear)
- Check if we can just take one or merge all storyboards, can you for example remove or add a certain segue in just one localization? (A missing segue or view controller could lead to generated code that crashes your app)

tomlokhorst:
If a missing segue in one localisation leads to a runtime crash, we could not generate that segue and print a warning instead. Similar to how R.swift now skips and warns about duplicate names.

mac-cain13:
I've looked a bit further into this issue and detecting localization variants seems quite easy, so we know what files are localized versions of each other.

I'm not completely sure if "just merging" is the way to go. I think often the optional return type will help us out, but not sure it will solve everything. I'll start out on a branch with merging all variants and then see what happens. :)

ikovalisko:
Thanks for ipdate! Please, let me know when you need testing this feature.

justinmakaila:
Is this fixed/merged? I recently added localization to my project and everything broke.

acecilia:
Happened to me the same, used Rswift all over the code and now that I am localizing I need to undo it.. :(

mac-cain13:
I'm very sorry to hear you ran into this while adding localization. I can see how that is a annoying thing to do. There is a local branch on my machine with this feature in it, but it still contains some bugs that I didn't had the time to look at it so no it is not yet merged/fixed.

On the flip side; It's a free and open source library, all development is done in the spare time of the maintainers without any payment received from any party. It's all love for the platform, community and library that drives us to build this. 
So while I appreciate that you make known to us that this is an important issue for you and other users it would also be nice not make it personal with phrases like _"because you don't have support for this"_.

I can't promise any timeline, but you certainly made us aware that this is a big thing for quite some users and I'll see when I can make some time to pickup the last work on that branch again. It won't be on the next few weeks though since I'm on a holiday first.

IlyaOstashkov:
I have the same annoying issue =(. There are 42 warnings like this:
warning: [R.swift] Skipping 738 strings in 'WordsPackHard' (en) because no swift identifier can be generated for all of these keys:

semireg:
[...]

PR #188, opened 2016-04-17
Generate R.string

tomlokhorst:
Implements https://github.com/mac-cain13/R.swift/issues/41 and is based on https://github.com/mac-cain13/R.swift/pull/176

Support for format specifiers has been added, based on code from [SwiftGen](https://github.com/AliSoftware/SwiftGen/blob/master/GenumKit/Parsers/StringsFileParser.swift). 

- [x] Support single Localizable.strings file used as a string constants file.
- [x] Support multiple Localizable.strings files in their respective .lproj directories.
- [x] Support printf style string formatting (format specifiers).
- [x] Create tests for format specifiers.
- [x] Handle mismatched patterns in format specifiers
- [x] Check if base translation should be special case.
- [x] Generate comment with base translation.
- [x] Create "LocalizableString" resources in R.swift.Libary, like for other resources.
- [x] Generate "lets" with LocalizableString resources.
- [x] Support multiple tables
- [x] Update documentation


tomlokhorst:
I've added a `StringResource` to the R.swift.Library and the generator.

This allows, for example, to print all translations of a key:

```swift
let string = R.string.settings.title

for locale in string.locales {
  if let path = NSBundle.mainBundle().pathForResource(locale, ofType: "lproj") {
    let bundle = NSBundle(path: path)

    if let localizedString = bundle?.localizedStringForKey(string.key, value: nil, table: string.tableName) {
      print("\(string.key) [\(locale)]: \(localizedString)")
    }
  }
}
```

This prints:
```
Title [Base]: Settings
Title [nl]: Instellingen
```

tomlokhorst:
I've split out the [stringsdict feature](https://github.com/mac-cain13/R.swift/pull/190) into a separate branch.

This pull request is feature complete and can be reviewed.

Documentation/Examples.md
```diff
@@ -71,6 +71,25 @@ label.textColor = R.color.appColors.textColor()
...
 There are some points to keep in mind when using Color palettes, see [About Colors](Colors.md)
 
+## Localized strings
+
+*Vanilla*
+```swift
+let welcomeMessage = NSLocalizedString("welcome.message", comment: "")
+let settingsTitle = NSLocalizedString("title", tableName: "Settings", comment: "")
+
+let welcomeName = String(format: NSLocalizedString("welcome.withName", comment: ""), locale: NSLocale.currentLocale(), "Alice")
```
mac-cain13: I think the example is better suited by using the `String(format:arguments:)` overload here. That one uses the default locale, prevents confusion about what the R.swift version below does with the locale.

R.swift/Generators/StringsGenerator.swift
```diff
@@ -0,0 +1,313 @@
...
+
+import Foundation
+
+struct StringsGenerator: Generator {
+  let externalStruct: Struct?
+  let internalStruct: Struct? = nil
+
+  init(localizableStrings: [LocalizableStrings]) {
+
+    let localized = localizableStrings.groupBy { $0.filename }
```
mac-cain13: Why do we first `groupBy` on filename and then again `groupBySwiftNames`, this `localized` variable seems to be never used again. So maybe this can be done all at once?

R.swift/ResourceTypes/LocalizableStrings.swift
```diff
@@ -0,0 +1,98 @@
...
+
+    return false
+  }
+}
+
+extension Locale: CustomStringConvertible {
+  var description: String {
+    switch self {
+    case .None:
+      return ""
```
mac-cain13: This feel quite tricky, when this is printed in for example a debug sentence you can get very confused by not seeing anything. Can we change this to something like `"None"` or do we need this to be an empty string? If so we should consider not using `CustomStringConvertible`, but another property/method.

R.swift/ResourceTypes/LocalizableStrings.swift
```diff
@@ -0,0 +1,98 @@
...
+
+  init(url: NSURL) throws {
+    try LocalizableStrings.throwIfUnsupportedExtension(url.pathExtension)
+
+    guard let filename = url.filename else {
+      throw ResourceParsingError.ParsingFailed("Couldn't extract filename without extension from URL: \(url)")
+    }
+
+    // Get locale from url (second to last component)
+    var locale = Locale.None
```
mac-cain13: I feel a `let locale: Locale` that is set to `.None` in the `else` branch is nicer, since it's immutable and compiletime checks that we don't forget to set it to the correct value. Maybe even an custom initializer or the `rawValue` initializer could help us out?!

R.swift/Generators/StringsGenerator.swift
```diff
@@ -0,0 +1,313 @@
...
+      type: Type(module: .Host, name: name),
+      implements: [],
+      typealiasses: [],
+      properties: params.map(StringsGenerator.stringLet),
+      functions: params.map(StringsGenerator.stringFunction),
+      structs: []
+    )
+  }
+
+  // Ahem, this code is a bit of a mess. It might need cleaning up... ;-)
```
mac-cain13: @tomlokhorst Could we take a look together at what a good cleanup approach would be?

R.swift/Generators/StringsGenerator.swift
```diff
@@ -0,0 +1,313 @@
...
+  let values: [(Locale, String)]
+
+  var localizedString: String {
+    let escapedKey = key.escapedStringLiteral
+
+    if tableName == "Localizable" {
+      return "NSLocalizedString(\"\(escapedKey)\", comment: \"\")"
+    }
+    else {
+      return "NSLocalizedString(\"\(escapedKey)\", tableName: \"\(tableName)\", comment: \"\")"
```
mac-cain13: Wondering if we should use a variant of localized strings that uses an explicit bundle like we do for images so it will also work with libraries that are localized.

Documentation/Examples.md
tomlokhorst: `String(format:arguments:)` uses the system locale, which is different from the current locale.

The [current locale](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSLocale_Class/index.html#//apple_ref/occ/clm/NSLocale/currentLocale):
> The logical locale for the current user. The locale is formed from the settings for the current user’s chosen system locale overlaid with any custom settings the user has specified.

Which is why it use used when formatting numbers and pluralisations.

Also this example matches the code that is generated: https://github.com/mac-cain13/R.swift/pull/188/files#diff-c67e8c23064b37ffa3ce51001930a68fR242

R.swift/Generators/StringsGenerator.swift
tomlokhorst: The first `groupBy` collects files from different locales, these all share the same filename.

The `groupBySwiftNames` is used to check for duplicates and empty swift identifiers.
This is only done _per_ filename group.

This code should be extracted and done somewhere else in a generic method. Related:
- [Validation methods don't take duplicate identifiers into account](https://github.com/mac-cain13/R.swift/issues/175)
- [Add support for localised storyboards and nibs](https://github.com/mac-cain13/R.swift/issues/155)
- [Refactor warnings so that it's not an sideeffect anymore](https://github.com/mac-cain13/R.swift/issues/136)

R.swift/Generators/StringsGenerator.swift
tomlokhorst: Sure, although most of the code in this method is related generating warnings. This should be cleanup as part of:
- [Add support for localised storyboards and nibs](https://github.com/mac-cain13/R.swift/issues/155)
- [Refactor warnings so that it's not an sideeffect anymore](https://github.com/mac-cain13/R.swift/issues/136)

R.swift/Generators/StringsGenerator.swift
tomlokhorst: This would be a nice improvement. It isn't very straightforward, as it effectively involves reimplementing `NSLocalizedString`. 

Since the generated `let` values in `R.string` can already be used to do this in user code, I think we can defer this feature to later.

Documentation/Examples.md
mac-cain13: That constructor states that it formats
> according to the user’s default locale

I think that is the same as the current locale, but it's not a big thing. Let's keep it as is for now.

R.swift/Generators/StringsGenerator.swift
mac-cain13: Check!

R.swift/Generators/StringsGenerator.swift
mac-cain13: Agreed that it should be fixed when we take up that warnings issue. I'll add that to the comment.

R.swift/Generators/StringsGenerator.swift
mac-cain13: Agreed, let's defer that improvement.

mac-cain13:
Alright, I say we'll merge this in and release a beta that we can put in some project to give it a try in the real world!

mac-cain13:
@tomlokhorst and @renrawnalon thanks a million for your work on this feature! It's awesome to have it in R.swift and I'm sure a lot of users will enjoy this! 👍

Issue #191, opened 2016-04-26
R.swift fails at parsing some files


Tvede-dk:
Problem from the outer world: some nibs (could be other files) are not parsed correctly (how i first noticed it)

Real problem: 
the resolved urls (resource urls) sometimes contains the path appended twice, like:
(real correct path): Users/kasper/some-app/SomeApp/SomeFile.xib 
(R.swift tries to open): Users/kasper/some-app/SomeApp/SomeApp/SomeFile.xib

Thoughts:
- first of, mabye there should be some error handling in the xml parsing, and returning a warning.
- second, the error handling should try to detect files that cannot be opened (eg just feed the parser a nsdata, since the parser do not tell about missing / invalid files). like so:

`let data = NSData(contentsOfURL: url.filePathURL!) `
`//if data == nil then the file could not be opened`
` let parser =  NSXMLParser(data: NSData(contentsOfURL: url)!)`

Error:
Somewhere in the resolution of urls,   (called from  Main.swift L 25 ) appends part of the path twice. 
The error is inside of PBXObject.
The error is whenever a group is relative to the project rather than the group, which means that if i have a group which is relative to a project rather than the group, the code would simply append the path one more time, which creates this odd behavior.

Workaround: change all paths to be relative to the group.

Solution:
inside of XCProjectFile.Paths handle files that are relative to the project.

mac-cain13:
Thanks for reporting this issue in such detail! Have to take a better look at it, but we'll fix it asap.

Tvede-dk:
I will try to see if i cannot come up with a PR for a solution, but i cannot guarantee a time frame. :)

Tvede-dk:
I have been able to create a solution for the problem, but that is inside of xcode.swift, and hence I do not know if the PR should be on this version of xcode.swift or the upstream ? 
(the code is at https://github.com/Tvede-dk/Xcode.swift) in the latest commit.

mac-cain13:
If you create the PR against upstream I'll ask @tomlokhorst to merge it in and we then can update R.swift to the latest Xcode.swift.

Would this fix in Xcode.swift mean that your previous PR #192 doesn't have to be merged anymore? Or is that still needed?

Tvede-dk:
[...]

PR #192, opened 2016-04-26
Adding error / warning messages to parsing stage when error(s) occur.

Tvede-dk:
Regarding #191 thoughts / recommendation (to display errors while parsing ): 
This does not solve #191 but rather makes it possible to see that R.swift failed parsing a particular file (or multiple). this can demystify some scenarios, where the generated R file would not contain the expected, yet there are no warning or errors while building.


mac-cain13:
Seems like you have another code formatting setting than that this project uses. Could you update the PR so the diff only contains your real changes? This makes it quite hard to review. :)

mac-cain13:
I like the idea you have about catching the errors that could happen, but this way doesn't feel totally right.

I'll merge this and adjust it so that:

- The Nib and Storyboard generator are both adjusted
- We don't use extra variables and load the data ourselves, but check if the constructor succeeded and check if the parsing succeeded.

PR #213, opened 2016-05-14
watchOS support for R.swift 🎉

tomasharkema:
At least, the bare minimum. I will spend time the coming days to implement more features, but I wanted to let you know my progress.

- [x] Colors
- [x] Fonts (won't think put effort in this one, since Fonts are not supported)
- [x] Images
- [x] ResourceFile
- [ ] Segue's (for nextPage in controller)
- [ ] Storyboard (will become very tricky)
- [x] Strings

And new for watchOS

- [ ] Notification Dynamic Identifier

⚠️ This contains changes on both Xcode.swift and R.swift.Library. Those PR's need to get merged first.
- https://github.com/tomlokhorst/Xcode.swift/pull/15
- https://github.com/mac-cain13/R.swift.Library/pull/9


mac-cain13:
Awesome you're working on this! I'll do a little review of the PR and let you know any concerns/questions I have. :)

.gitignore
```diff
@@ -3,6 +3,7 @@ xcuserdata
 *.xccheckout
 ResourceApp/rswift
 R.generated.swift
+R.watchOS.generated.swift
```
mac-cain13: Why do we generate a seperate file for watchOS, is this necessary? The watch is a seperate target if I'm correct, just like tvOS and iOS are separate. I think it would be possible (and nicer) to generate a `R.generated.swift` file for the watchOS target, it has also needs it's own setup in that case.

But correct me if I'm wrong and this isn't possible.

tomasharkema: Good point. When I built this, my fork was not up to date with `master`, so tvOS was not yet included, so I took my own approach. I'll change it.

R.swift/ResourceTypes/Xcodeproj.swift
```diff
@@ -24,6 +49,32 @@ struct Xcodeproj: WhiteListedExtensionsResourceType {
...
+
+    switch (configs.first?.buildSettings.SDKROOT, configs.first?.buildSettings.IPHONEOS_DEPLOYMENT_TARGET) {
+    case let (sdkRoot?, _):
+      guard let targetType = TargetType.fromSdkRoot(sdkRoot) else {
+        throw ResourceParsingError.OsNotSupported(sdkRoot)
+      }
+
+      return targetType
+
+    case (_, _?):
```
mac-cain13: I think `.Some` is clearer to read here.

R.swift/Generators/ResourceGenerator.swift
```diff
@@ -31,19 +31,33 @@ private struct GeneratorResults {
...
       ColorGenerator(colorPalettes: resources.colors),
       FontGenerator(fonts: resources.fonts),
       SegueGenerator(storyboards: resources.storyboards),
       StoryboardGenerator(storyboards: resources.storyboards),
       NibGenerator(nibs: resources.nibs),
       ReuseIdentifierGenerator(reusables: resources.reusables),
       ResourceFileGenerator(resourceFiles: resources.resourceFiles),
-      StringsGenerator(localizableStrings: resources.localizableStrings),
+      StringsGenerator(localizableStrings: resources.localizableStrings)]
+  case .watchOS:
```
mac-cain13: I would reserve this switch only for thing where it's absolutely necessary to differentiate between watchOS and iOS. For example; Images look exactly the same, so the generated doesn't know whether it must generate an image for watchOS or iOS, but the code must be different.

For a storyboard for example it doesn't matter, the generator simply won't encounter iOS classes in a watchOS storyboard, so it will automatically generate the correct imports and use te correct classes. Should never be a problem, so we don't have to use seperate code paths for them.

mac-cain13:
Some notes in general:

- It's good to have a (semi) complete list of watchOS features to support, we don't have to implement them all in this PR. The more complex and lesser used ones can move to separate issues and be implemented later.
- The lesser exception in code we have for iOS/watchOS the better, this will keep the codebase maintainable. (But I can see we need some exceptions here and there.)

Also; agreed on fonts, watchOS doesn't support custom fonts, so there is no use in supporting them.

mac-cain13:
@tomasharkema Do you still want to finish this PR?

fruitcoder:
Any news on watchOS support? I have a framework that encapsulates common logic from my targets and I would like to use R.swift to generate a public interface for every target (iOS, watchOS, tvOS) to access the localized Strings.

mac-cain13:
@tomlokhorst mentioned to me that he has started on this a few days ago, but he just got started so no idea when this will be in a testable state. :)

fruitcoder:
Okay, thanks for the info!

flaviobittenca:
Any perspective to watchOS support? :)

mac-cain13:
No one is currently working on this, so no not really. This PR is quite old and outdated with the current codebase so I think it would be easier to just start over if anyone would like to create this.

It's just that I don't work on any watchOS apps atm, so I can't afford to work on this feature professionally and in my spare time I have other features I rather work on than watchOS. :(

PR #221, opened 2016-05-26
Close view controller on scene, because we need the extra objects

zhigang1992:
screen shot 2016-05-26 at 2 17 35 pm

When the segue is not **exactly** in the view controller, but in the scene, we should generate those segues as well.

This might not be a very good fix, but it seems to be the easiest.


tomlokhorst:
Possibly related to https://github.com/mac-cain13/R.swift/issues/223, https://github.com/mac-cain13/R.swift/issues/224.

mac-cain13:
Thanks for creating a PR! Seems that #249 will fix this also and I'll merge that one.

Issue #232, opened 2016-06-24
Feature request: Include .strings file comment in generated type documentation


nolanw:
For example, if I have in my `.strings` file:

```strings
"search-placeholder" = "Search %@"; /* Parameter is name of category. */
```
it'd be awesome to have the generated function and property documentation do something like
```swift
/// Base translation: Search %@
/// 
/// Locales: Base
///
/// - Note: Parameter is name of category.
```
so you get the localizer note right at the code use site. It's a helpful reminder of what the string is meant for and what the parameters are. It also makes bad/outdated localizer notes a bit more visible so maybe they can be fixed.

I believe `genstrings` puts the comment as I've done in my example `.strings` file, so that's probably as close as we'll get to a standard for comment placement.

Loving R.swift btw, thanks for making it!

mac-cain13:
@tomlokhorst Do you know if this is easy to do? Are the comments available to us when we parse the file?

Sounds like a nice feature!

tomlokhorst:
Comments were part of the original design. But they fell off, because the default parsing of a plist file to a NSDictionary doesn't expose them. 

I still like the feature, though!

To parse the comments we'd need to create a plist parser that exposes them. Or find a third-party parser that works, I haven't looked for one.

nolanw:
I had no idea you could parse `.strings` as a plist using `NSDictionary`! I can't decide if that's really cool or slightly horrifying ;) But now that you mention it, if you tack curly braces around a `.strings` file you get an ASCII plist.

I'd assumed there was a lil parser for this already in R.swift. Using `NSDictionary` makes this a larger feature request than I thought it would be. Sorry about that.

Some random notes for anyone who takes this on (maybe me), or maybe for evaluating an existing parser to bring in:

* The parser needs to consider escapes (e.g. `\"` doesn't mark the end of the key/value), which I've seen done badly by lots of code that ostensibly parses `.strings` files.
* There's also `\Uxxxx` escapes to maybe consider. Surrogate pairs are accepted here by `NSDictionary` so watch out for those.
* I may be wrong about where to put the comment; the [Resource Programming Guide](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html#//apple_ref/doc/uid/10000051i-CH6-SW10) says the convention is to put the comment on the line before the string it's describing.

nolanw:
More notes to myself/whoever:

```strings
/*
/*
*/
[...]

PR #235, opened 2016-07-01
Parse localizable strings comments

nolanw:
This is an attempt to start implementing #232.

* [x] Switch to a regex for parsing `.strings` files so we can capture comment.
* [x] Warn when a duplicate key is encountered in a `.strings` file (previously the last value was silently chosen the winner).
* [ ] Include comment that precedes `.strings` file entry, if present, in `R.generated.swift`.

I'm a bit stuck on the last part, where we carry the parsed strings comment all the way through to `R.generated.swift`. I'd appreciate some guidance! (Or if someone wants to just take it from here, go for it!)

### DIFF
--- a/R.swift/Generators/StringsGenerator.swift
+++ b/R.swift/Generators/StringsGenerator.swift
@@ -60,7 +60,7 @@ struct StringsGenerator: Generator {
   // Maybe when we pick up this issue: https://github.com/mac-cain13/R.swift/issues/136
   private static func computeParams(filename: String, strings: [LocalizableStrings]) -> [StringValues] {
 
-    var allParams: [String: [(Locale, String, [StringParam])]] = [:]
+    var allParams: [String: [(Locale, LocalizableStrings.Entry)]] = [:]
     let baseKeys: Set<String>?
     let bases = strings.filter { $0.locale.isBase }
     if bases.isEmpty {
@@ -95,10 +95,10 @@ struct StringsGenerator: Generator {
       for key in groupedKeys.uniques {
         if let entry = byKey[key] {
           if let _ = allParams[key] {
-            allParams[key]?.append((ls.locale, entry.val, entry.params))
+            allParams[key]?.append((ls.locale, entry))
           }
           else {
-            allParams[key] = [(ls.locale, entry.val, entry.params)]
+            allParams[key] = [(ls.locale, entry)]
           }
         }
       }
@@ -138,8 +138,8 @@ struct StringsGenerator: Generator {
       var params: [StringParam] = []
       var areCorrectFormatSpecifiers = true
 
-      for (locale, _, ps) in keyParams {
-        if ps.any({ $0.spec == FormatSpecifier.TopType }) {
+      for (locale, entry) in keyParams {
+        if entry.params.any({ $0.spec == FormatSpecifier.TopType }) {
           let name = locale.withFilename(filename)
           warn("Skipping string \(key) in \(name), not all format specifiers are consecutive")
 
@@ -149,8 +149,8 @@ struct StringsGenerator: Generator {
 
       if !areCorrectFormatSpecifiers { continue }
 
-      for (_, _, ps) in keyParams {
-        if let unified = params.unify(ps) {
+      for (_, entry) in keyParams {
+        if let unified = params.unify(entry.params) {
           params = unified
         }
         else {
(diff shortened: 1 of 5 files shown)


tomlokhorst:
Very cool! Thanks for this PR!

I quickly scrolled through this, I'll look in to it more deeply when I have some more time.

The warnings about duplicate identifiers should be part of the code generation step, not the parsing step. That's because multiple files can be merged together at code generation time. When parsing it is not yet known which keys will become duplicate unique identifiers.

In this case that means the `parseString` function should actually return a list of tuples (or structs) instead of the dictionary it returns now.

As for the rest of it; I'll take a closer look later, or maybe someone else will before that.

nolanw:
The warning I added is meant for duplicate keys in a single file; I thought there was a todo somewhere about maybe adding a warning for that. NSPropertyListSerialization just silently takes the last of the duplicates.

tomlokhorst:
Standard R.swift behaviour for duplicate keys (single file or not), is to skip generating code for that key and only print a warning.

An extra feature would be to extend that behaviour to strings files. That means collecting all keys (including duplicates within a single file), and have the code generator print the warnings (which it already does).

This also allows for extensibility in the future, where some other process between the parser and the code generator does something with all the keys. Like show them to the user or filter them based on some heuristic. 

Again, this is an additional feature that R.swift currently doesn't have. But the reason for building a custom parser, instead of using the plist parser, is to enable additional features.

nolanw:
Here's a first attempt at the rest of this PR, getting the parsed `.strings` file entry comment into the generated documentation comment. No idea if this is the right way to do this, so feedback very welcome!

I removed the warning emitted during parsing (thanks @tomlokhorst for patiently explaining). We can add it to the appropriate place sometime later. [Here's the todo I was talking about that mentions this warning](https://github.com/mac-cain13/R.swift/blob/master/ResourceApp/ResourceApp/Strings/en.lproj/Localizable.strings#L9), if anyone's curious (shout out to fellow Nolan creating that file).

I've also modified the regex a bit to only consider a comment directly preceding (i.e. up to 1 line away from) a `.strings` file entry as associated with that entry. Previously the comment atop the file was getting attached to the first entry, and that's not helpful. Unfortunately (?) I couldn't figure out how to do this entirely with the regex.

nolanw:
It's been a month, is there anything I can do to move this PR along? Y'all won't hurt my feelings if you share what you don't like about it!

tomlokhorst:
Hey Nolan, sorry for leaving this PR hanging for over a month...

Due to both me and @mac-cain13 being busy with other things, R.swift maintenance is on a bit of a summer break at the moment. 

Unfortunately that means issues and pull requests start stacking up. Sorry about that.

The next thing I plan on doing for this PR is;
 - Test this branch agains my internal apps using `R.string.*`, to see if anything breaks.

Hopefully I'll find the time to do that in the next week or so.

nolanw:
No worries, enjoy the break!

tomlokhorst:
One thing that would be nice to add to this PR: Tests :-)

- [ ] In ResourceApp add a `comments.strings` which has different keys with comments
- [ ] In ResourceAppTests/StringTests add `XCTAssertEqual` to check if the comments are correctly parsed

tomlokhorst:
I just tested this on my biggest project using `R.string.*`.
The parser works great! It didn't skip anything, I did find a accidental duplicate key in my project.

I think these should be the test cases:

```
/* comment doesn't belong to any key */

"key1" = "One";

/* comment belongs to key2 */
"key2" = "Two";

"key3" = "Three"; /* comment belongs to key3 */
"key4" = "Four";

/* first part of comment for key5 */
"key5" = "Five"; /* second part of comment, should be merged */

/* big, multiline, comment
 * belonging to key6
 */
"key6" = "Six";
```

Also nice to have are single line comments:


```
// comment doesn't belong to any key

"key1" = "One";

// comment belongs to key2
"key2" = "Two";

"key3" = "Three"; // comment belongs to key3
"key4" = "Four";

// first part of comment for key5
"key5" = "Five"; // second part of comment, should be merged

// big, multiline, comment
// belonging to key6
"key6" = "Six";
```

nolanw:
Tests sound good! I have a question about how to implement them. Does it make sense to include a `comment` property on `StringResource`? I don't anticipate these comments being useful to have in code, I had expected them just to be in the Quick Help. However that makes them more difficult to test. Any suggestions for how to test the generated comments?

tomlokhorst:
Yes, I do think it makes sense to add a `comment` property. 

The `StringResource` exists to expose all values that R.swift has collected to third party libraries to build against. I don't think there's a reason to exclude the comments from that.

Indeed, the tests can also use the `comment: String?` property.

nolanw:
Cool, that makes sense. I've added the `comment` property in my fork of `R.swift.Library` and updated the submodule here; I'll open a PR over there once this one's in good shape.

Thanks for so patiently explaining like every part of this project to me as I go.

I started messing with the regex to consider `//`-comments and both leading and trailing comments. I got pretty lost. I tried again with an `NSScanner` approach and despite the inflation in lines of code it seems clearer to me. There are currently failing strings tests but they're cosmetic; I need to nail down what the merged comments should look like and then test that they look like that. I'll look into the other test failures too, but that's all for today.

tomlokhorst:
The `comment` property is now available in the `StringResource` in the master branch of R.swift.Library.

tomlokhorst:
I've created a Swift 3 version of this PR: https://github.com/mac-cain13/R.swift/pull/283

PR #244, opened 2016-07-30
Refactor to SwiftIdentifier

tomlokhorst:
Implements first part of https://github.com/mac-cain13/R.swift/issues/136#issuecomment-222360068

Future work:

 - Add localized resources warnings
 - Refactor warnings so that it's not an sideeffect anymore

Update this pattern:

```swift
let groupedResourceFiles = resourceFiles.groupBySwiftIdentifiers { $0.fullname }
groupedResourceFiles.printWarningsForDuplicatesAndEmpties(source: "resource file", result: "file")

// groupedResourceFiles.uniques.map(stuff)
```

To the following:

```swift
let groupedResourceFiles = resourceFiles.groupBySwiftIdentifiers(
  identifier: { $0.fullname }
  source: "resource file",
  result: "file"
)

// groupedResourceFiles.map(stuff)
```


tomlokhorst:
Related bug I found while testing https://github.com/mac-cain13/R.swift/pull/235:

If there are duplicate keys in localized string files, a warning is printed, but the key isn't skipped for other localizations. This should be fixed when these two PRs are merged

- [ ] Duplicates in one locale should also skip other locales

Example where keys aren't correctly skipped:

```
// Localized.strings (en)
one = One;
one = One;

// Localized.strings (nl)
one = Een;
```

PR #255, opened 2016-08-24
Don't generate LocalizableStrings for localized nibs

tomlokhorst:


tomlokhorst:
Thinking about this some more, this may not be the correct solution for this problem.

Maybe the .strings file related to the .nib or .storyboard can also be used by the user to add other strings? If so, we shouldn't exclude the complete file, only the keys used in the nibs/storyboards.

Issue #252, opened 2016-08-15
Export for Localization generates unwanted translation units


CosynPa:
For example when I localize launchScreen.storyboard which has a label whose ID is "cZB-if-HYR", R.generated.swift will has the following code:

``` swift

  /// This `R.string` struct is generated, and contains static references to 2 localization tables.
  struct string {
    /// This `R.string.launchScreen` struct is generated, and contains static references to 1 localization keys.
    struct launchScreen {
      /// zh-Hans translation: Label
      /// 
      /// Locales: zh-Hans
      static let cZBIfHYRText = StringResource(key: "cZB-if-HYR.text", tableName: "LaunchScreen", locales: ["zh-Hans"])

      /// zh-Hans translation: Label
      /// 
      /// Locales: zh-Hans
      static func cZBIfHYRText(_: Void) -> String {
        return NSLocalizedString("cZB-if-HYR.text", tableName: "LaunchScreen", comment: "")
      }

      private init() {}
    }

// ...

```

This line `NSLocalizedString("cZB-if-HYR.text", tableName: "LaunchScreen", comment: "")` will cause XCode "export for localization" to generate unwanted translation unit in the xliff file: 

``` xml
<trans-unit id="cZB-if-HYR.text">
        <source>cZB-if-HYR.text</source>
        <note>No comment provided by engineer.</note>
</trans-unit>
```

And the real translation unit for this label is already generated. And also the function `cZBIfHYRText` in `R.genrated.swift` is not human-readable, I think it should be removed.

hefgi:
Any news on this issue ? Thanks

mac-cain13:
@tomlokhorst would it be possible to cherry-pick that commit that fixed this?

tomlokhorst:
This was never properly fixed.

The workaround of excluding files that are named after nibs works, but that might also exclude perfectly valid user-defined translations.

The correct solution would be: 
 - Scan all storyboard and nibs
 - Find the id's for all localisable items (I assume these are all the views items under `<subviews>`, but there might be more)
 - Exclude these ids from being generated in R.string.*

I think this issue is purely cosmetic. The currently generated code is not needed, but isn't harmful either. Therefor we should go for the proper solution instead of a quick workaround that _is_ potentially harmful.

tomlokhorst:
I don't know why I closed this issue before, it wasn't fixed in 7.0.0.

[...]

PR #268, opened 2016-09-29
Feature/filter generated translation

mac-cain13:
@tomlokhorst Do we need to merge this? It was in the history, but not on a branch and is not merged yet. So I'm wondering if this should be merged or if it is not a good fix and should be ignored.


mac-cain13:
Discussed with @tomlokhorst and this is not the correct fix for #252 so this should be ignored.

PR #275, opened 2016-10-09
Lowercase the full uppercase prefix of a name

### LINKED ISSUE #179: Class prefix detection
I have a class WPMenuViewController. R.swift generates a struct with a camel case name wPMenuViewController. Maybe this should just stay as WPMenuViewController or wp_MenuViewController. I have my class prefix set in my project.

mac-cain13:
Fixes #179 

`NSSomeThing` was converted to `nSSomeThing` and will now be converted to `nsSomeThing` instead. Xcode will automatically create fix-its for this scenario.

Note; If you had `nSSomeThing` and `nsSomeThing` as identifiers in your project this will now break since they will become duplicate identifiers. You will get a warning about this.

I think strictly this should be in a major release because it is a breaking change. But it's not that big, a few users will get fix-its and some other will have te rename a few identifiers.

@tomlokhorst What do you think? Should this wait for a major release or should we just merge it in.


tomlokhorst:
Since this breaks existing code, this should definitely be a major version release.

Unfortunately I can't test this on any of my bigger projects, because they're all still Swift 2.3...

mac-cain13:
Okay, I think this should be behind a feature flag and then become the default in the next major release.

PR #281, opened 2016-10-18
Support skipping files

shiraji:
This is a first draft of skipping resource files.
## Not Implemented, yet

At this moment, following features are not implemented.
- wildcard
- skipping all files in folder
- must not start with "/" (.gitignore ignores first slashes)
- negation ("!" symbol at the beginning of the line)
## Sample skip file

Same as .gitignore file, developers should specify relative path from --rswiftignore file. (I don't think it should accept using "..")

```
# filter Colors.jpg
ResourceApp/Images/Colors.jpg
# filter GdyBkltter1911.ttf
ResourceApp/Fonts/GdyBkltter1911.ttf
```
## Other comments

I cherry-pick your commits but some of them conflict with master. So, you better check all of the commits listed here.

Let me know if you want me to squash my commits.

Thanks, in advance!


shiraji:
Could you give me some clue how to fix this problem? or how to set up CI?

mac-cain13:
Seems like Bitrise tries to checkout your commit on the wrong repo or something. No worries about that, I'll look into it later.

I can only authorize 2 persons atm so I cant easily let you in Bitrise. But if you build and run all tests locally it should give the same result as Bitrise does.

shiraji:
Thanks! As long as you can check my code, I'm ok.

shiraji:
Do you want me to support wildcard or specify folder features in this PR? (I'm not sure how long does it take...)

ResourceAppTest's `testWarningsAreLogged` fails even in master branch. Other than that since I didn't add any test cases, this PR passes all test cases.

R.swift/input.swift
```diff
@@ -177,6 +186,12 @@ struct CallInformation {
 
       productModuleName = try getFirstArgumentForOption(productModuleNameOption, environment["PRODUCT_MODULE_NAME"])
 
+      if let rswiftignorePath = options[rswiftignore]?.first {
+        rswiftignoreURL = URL(fileURLWithPath: rswiftignorePath)
+      } else {
+        rswiftignoreURL = nil
+      }
```
mac-cain13: This fallback is exactly what `getFirstArgumentForOption` will do right? So I don't see why we shouldn't use that here.

R.swift/main.swift
```diff
@@ -13,13 +13,28 @@ let IndentationString = "  "
 
 let ResourceFilename = "R.generated.swift"
 
+private func isValidURL(ignoreFile : IgnoreFile?, url: NSURL) -> Bool {
```
mac-cain13: I think it would be simpler to create a `IgnoreFile.init()` without any arguments that creates an empty ignore file. Then all the logic can stay in one place and there are less exception.

mac-cain13:
Looks good, thanks for your effort so far, I think the most important feature to have is wildcard support. The most common use cases imo are:

- Ignore all files of a certain type: `**.ttf` should ignore any `ttf` file regardless folder depth
- Ignore a whole directory: `some/folder` should ignore anything in that folder including subfolders
- Ignore certain files in a folder: `some/folder/*.jpg` should ignore all jpg in that folder
- Ignore certain files in a folder: `some/folder/**.jpg` should ignore all jpg in that folder and subfolders

I think that it would be most practical to call out to the `glob` c-function. That would have all these features without we having to implement that ourselves. What do you think?

shiraji:
I like your idea comparing with reinventing the wheel. I will work on that.

shiraji:
So, I integrated `glob` and notice it does not support `**`. I think Python or other language `glob` supports `**` but not C. 

22c9ad991c839e78cb85ef493aba2081da0a36d9 only support the following feature using `glob`.

* Ignore certain files in a folder: `some/folder/*.jpg` should ignore all jpg in that folder

By the way, does `gitignore` support the following feature?

* Ignore certain files in a folder: `some/folder/**.jpg` should ignore all jpg in that folder and subfolders

mac-cain13:
Cool! Thanks for your work on this.

Git does support it as [stated in the documentation](https://git-scm.com/docs/gitignore#_pattern_format), but it seems they don't use the C implementation of the glob function. I looked into it and found [the commit](https://github.com/git/git/commit/237ec6e40d4fd1a0190c4ffde6d18278abc5853a) where they implement this feature. There you see that they use [wildmatch](https://github.com/git/git/blob/master/wildmatch.c) to implement it.

Along the way I found this [C library called wildmatch](https://github.com/davvid/wildmatch) with the same name (it's not the same code as git uses, but it has about the same feature set). We could look into it, if we could include this in the R.swift project we could use that implementation of `wildmatch` and also support the `**` feature.

What do you think?

shiraji:
That works. I will take a look.

I don't read code, yet but I found Swift version of glob with `**` support in gist https://gist.github.com/efirestone/ce01ae109e08772647eb061b3bb387c3

mac-cain13:
That also looks promissing! Do you have some time to give one of these options a try? :)

shiraji:
Of course! One thing I'm concern about is I'm not good at licensing. If my pull request violate license, please let me know.

mac-cain13:
Apache license seems okay, especially if we just leave that code alone. I'll also add a statement in the Readme if we depend on an other library.

shiraji:
The Swift version of Glob does not work well for `**.strings` or `*.strings`. It works if the patten is `**/*.strings`. I will try wildmatch

shiraji:
Could you help me integrating C library? Here is my working branch https://github.com/shiraji/R.swift/tree/skip-file-wildmatch

mac-cain13:
Looked into it, turns out the bridging header wasn't configured correctly, this PR fixes it: https://github.com/shiraji/R.swift/pull/1

Would be nice if `wildcard` does what we want! If it turns out that it's hard to use or also doesn't support the exact matching we need I think the Swift gist version could also be okay. `**/*.strings` isn't a bad syntax perse, but if we can have more gitignore style syntax that would be nice!

shiraji:
Thanks for your pull request! It compiles now. However, the result is the same as Swift's version. Since this library is written by Swift, I think the Swift gist version would be better. What do you think?

mac-cain13:
Awesome! Thanks for putting time into this and figuring this our.

Also agreed, having C code in the project is not ideal, would be nicer to use the Swift version if they have nearly the same features.

shiraji:
Add Glob.swift 🚀 

## Current features

* Ignore certail file: `some/folder/foo.strings` should ignore the exact file
* Ignore certain files in a folder: `some/folder/*.jpg` should ignore all jpg in that folder
* Ignore certain files in a folder: `some/folder/*` should ignore all files in that folder
* Ignore certain files in a folder: `some/**/*.jpg` should ignore all jpg in sub directory of `some`.
* Ignore a whole directory: `some/folder/*` should ignore anything in that folder including subfolders

## Missing features

* Ignore all files of a certain type: `**.ttf` should ignore any ttf file regardless folder depth
* Ignore a whole directory: `some/folder` should ignore anything in that folder including subfolders

I thought just use `hasPrefix` instead of using `==` but this leads filtering wrong files.

I guess if we want to those features, we need to change Glob.swift

Now I start wonder how I can test this code...Do you think I should integrate Glob.swift's test, too?

mac-cain13:
Awesome! Thanks for the work so far.

Some questions that come to mind:

- You describe "ignore a whole directory including subfolders" and "ignore all files in that folder" both as `some/folder/*`. I think there is a typo in there somewhere?
- "Missing feature" `**.ttf` doesn't work, but can you instead do `**/*.ttf`? That would be a good workaround I think.

On the testing, yes I think it is good to put the glob tests als into R.swift and probably we should write some tests for `IgnoreFile.swift` also in R.swift itself. The `ResourceApp` can have an ignore file for a few "end-to-end" tests to see if that is working.

Note to myself, we shouldn't forget to write some documentation about this feature.

shiraji:
OK, update it! I forget to mention `**/*.ttf` feature. 

## Current features

* Ignore certail file: `some/folder/foo.strings` should ignore the exact file
* Ignore certain files in a folder: `some/folder/*.jpg` should ignore all jpg in that folder
* Ignore certain files in a folder: `some/**/*.jpg` should ignore all jpg in directory of `some`
* Ignore a whole directory: `some/folder/*` should ignore anything in that folder including subfolders
* Ignore all files of a certain type: `**/*.ttf` should ignore any ttf file regardless folder depth

## Missing features

* Ignore all files of a certain type: `**.ttf` should ignore any ttf file regardless folder depth
* Ignore a whole directory: `some/folder` should ignore anything in that folder including subfolders

Diff!

```diff
## Current features

* Ignore certail file: `some/folder/foo.strings` should ignore the exact file
* Ignore certain files in a folder: `some/folder/*.jpg` should ignore all jpg in that folder
-* Ignore certain files in a folder: `some/folder/*` should ignore all files in that folder
-* Ignore certain files in a folder: `some/**/*.jpg` should ignore all jpg in sub directory of `some`.
+* Ignore certain files in a folder: `some/**/*.jpg` should ignore all jpg in directory of `some`
* Ignore a whole directory: `some/folder/*` should ignore anything in that folder including subfolders
+ * Ignore all files of a certain type: `**/*.ttf` should ignore any ttf file regardless folder depth

## Missing features

* Ignore all files of a certain type: `**.ttf` should ignore any ttf file regardless folder depth
* Ignore a whole directory: `some/folder` should ignore anything in that folder including subfolders
```

shiraji:
OK, I added test cases for Glob.swift. This actually helpful writing documentation.

I have questions about test cases. I'm trying to write test cases that the resource is missing. 

ignore file

```
# Filtering single file
ResourceApp/Images/Sky.tiff
```

Testing

```swift
let imageMirror = Mirror(reflecting: R.image.self)
print(imageMirror.children.count) // show 0.
``` 

I thought using Mirror is the only way to test it. Do you have any suggestion?

By the way, I'm having hard time to creating test cases for IgnoreFile.swift. If I add test cases for ResourceApp's ignore file, is it OK? or do you have any idea how to test IgnoreFile.swift?

shiraji:
Could you help me figure out how to test this, @mac-cain13 ?

mac-cain13:
For sure! I'm crazy busy atm, but as soon as I have some time I'll dive into this PR. I think we're pretty close to a first version of this feature that we can merge.

Thanks again for your time!

shiraji:
Good to hear that! I will leave this PR until you give me feedback.

Sorry for borthing you many times.

2016年10月31日(月) 16:37 Mathijs Kadijk <notifications@github.com>:

mac-cain13:
Finally took some time to review/test/merge this PR! It's in master now and will be in the next release. Main thing I want to do is add a test and run it in a real life project before releasing a stable version.

Thanks again for all the time you put into this!

shiraji:
@mac-cain13 Nice! Thanks a lot.

PR #283, opened 2016-10-24
Localizable strings comments

tomlokhorst:
Implements https://github.com/mac-cain13/R.swift/issues/232, continuation of https://github.com/mac-cain13/R.swift/pull/235

I've merged the current master branch into @nolanw's code, making it all Swift 3.

Todo:

 - [ ] Multiline or multi-part comments should be separated by `\n`
 - [ ] Comments should be printed in generated code
 - [ ] All tests should succeed


hujunfeng:
Any plan to continue and finish this work? The last update of this MR was 1.5 years ago.

If no one is going to pick it up anytime soon, I would like to give it try.

nolanw:
No plans from me, I ran outta steam.

tomlokhorst:
We sort of stopped with this PR when @mac-cain13 and I realised we didn't want to replace your Strings parser, but only augment it.

So only build a Strings Comment parser, and combine that with the existing Strings parser. That way, existing file won't break if the new parser isn't perfect.

Anyway, that was the idea, but we never did anything more with it.

anton-kaliuzhnyi-altexsoft:
@tomlokhorst Are we going to see #232 in production any time soon, please?

PR #296, opened 2017-01-04
Pass in a bundle when validating images in nibs.

Noobish1:
when validating images in nibs, pass in R.hostingBundle as the bundle so the main bundle isn't used by default


Noobish1:
Bitrise, not sure why it's failing as I can't access it.

Noobish1:
@mac-cain13 anything that can be done with this?

tomlokhorst:
Hi @Noobish1, sorry for not responding sooner.
Bitrise is failing because it can't find the commit in this repository, you can ignore that.

Can you add an example/test for this in the ResourceApp tests? I'm unsure how to load an image from a different bundle in a Storyboard.

mac-cain13:
@tomlokhorst I assume that if the Xib/Storyboard is in bundle X that the images referenced in that Xib/Storyboard will be looked up also in bundle X. Even if that is not the main bundle, but I don't think I can check that easily in our current test setup.

This change seems valid since we need to load normal UIImages the same way. I'll merge this in.

Noobish1:
Sorry for not replying to this earlier. We have a bunch of swift frameworks in our apps and we have R.swift in a couple of those frameworks, so if images are located inside those frameworks R.swift doesn't validate them correctly because it looks in main instead of our framework.

mac-cain13:
Alright, so that should be fixed in the next release with this PR merged! 😃

PR #311, opened 2017-03-28
Use Swift Package Manager

tomlokhorst:
(This PR was previously just for the XcodeEdit update)

- [x] Use SPM for XcodeEdit
- [x] Remove R.swift.Library submodule
- [x] Replace OptionKit submodule with SPM Commander
- [x] Use SPM to build `rswift` itself
- [x] Get bitrise to work with SPM
- [x] Update release scripts


tomlokhorst:
Since it isn't 2014 anymore, maybe we should move away from git submodules; They're annoying.
I'll update this PR to use Swift Package Manager instead, and throw away all the submodules.

tomlokhorst:
Work complete. R.swift can now be build with SPM.

Sources/rswift/Rswift.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by Tom Lokhorst on 2017-04-22.
+//  From: https://github.com/mac-cain13/R.swift
+//  License: MIT License
+//
+
+import Foundation
+
+struct Rswift {
+  static let version = "3.9.0"
```
tomlokhorst: Version still needs to be manually set.

Sources/RswiftCore/Util/UtilExtensions.swift
```diff
@@ -19,12 +19,29 @@ extension Array {
...
+
+    return groupedBy
   }
 }
 
+// extension Sequence where Iterator.Element : CustomStringConvertible {
+//   func joined(separator: String) -> String {
+//     return map { $0.description }.joined(separator: separator)
+//   }
+// }
```
mac-cain13: Can we just remove this function?

Sources/RswiftCore/SwiftTypes/Module.swift
```diff
@@ -9,23 +9,23 @@
 
 import Foundation
 
-enum Module: ExpressibleByStringLiteral, CustomStringConvertible, Hashable {
+public enum Module: ExpressibleByStringLiteral, CustomStringConvertible, Hashable {
```
mac-cain13: @tomlokhorst Why do these enum + methods need to change to public?

Sources/rswift/Rswift.swift
mac-cain13: Yes, need to update the release script that bumps this version.

Sources/RswiftCore/SwiftTypes/Module.swift
tomlokhorst: The `rswift` executable now depends on the module `RswiftCore`. These need to be public to be used from the executable.

Sources/RswiftCore/Util/UtilExtensions.swift
tomlokhorst: Yes, that is beter. 😄

mac-cain13:
So, release script works, tests all run correctly with Fastlane and from Xcode on my local machine. Only Bitrise is not happy yet, it gives us a linking error when running the tests from Fastlane.

Fixing Bitrise is the only thing preventing me from merging this atm, will take a look if I can fix Bitrise tomorrow!

Issue #309, opened 2017-03-23
.clr generated from Zeplin not translated well in XCode


YotamGolomb:
When defining colors in Zeplin and exporting it to .clr file , the R.swift isn't creating the right colors.
e.g : i have created a color from RGB 77 217 180 which is 4DD9B4. it was translated to 43D4A5

tomlokhorst:
This may be related to https://github.com/mac-cain13/R.swift/issues/301

R.swift currently always assumes the colors defined in a .clr are in the sRGB color profile. 
If the .clr file is in a different color profile, R.swift converts the color incorrectly. This is a bug.

Do you know in what color profile the .clr file is that was exported from Zeplin?

YotamGolomb:
All i know is that the Color Representation of the project is automatic. As for the exporting option , it doesn't allow me to choose a profile

mac-cain13:
I've merged the PR that checks the colorspace so the next release of R.swift should at least warn you if the color space is not correct. Hope that at least makes this problem clearer.

@YotamGolomb could you share a problematic clr file so we can check if the color profile in your clr is indeed the problem?

mac-cain13:
Also; Support for P3 colors in CLR files would be nice!

mac-cain13:
I'm going to assume this is fixed with PR #312, P3 support has its own ticket now: https://github.com/mac-cain13/R.swift/issues/317

Issue #301, opened 2017-02-24
Colors Hex - Incorrect


xiaosongshu:
Would there be any reason my colors from my .clr file would appear in R.generated incorrectly? 

For example, in Sketch, I'm saving "#F2F6F7" in my .clr file and in R.generated, I have:

```
/// <span style='background-color: #EFF4F5; color: #100B0A; padding: 1px 3px;'>#EFF4F5</span> White smoke
      static let whiteSmoke = Rswift.ColorResource(name: "White smoke", red: 0.935933113098145, green: 0.955909192562103, blue: 0.960774302482605, alpha: 1.0)
``` 

`#EFF4F5` is VERY close to `#F2F6F7`, but not exact. This is happening with all of my colors. Am I doing something wrong?

tomlokhorst:
I'm unable to reproduce this in the quick test I just did.
If I create a .clr file from the system color picker with a `#F2F6F7` color, R.swift will interpret it as such.

How are you creating a .clr file from Sketch? I looked around and couldn't find a way to do that.

I did notice that R.swift currently always assumes a RGB color space, that might be incorrect on macOS and iOS 10 that support more color spaces. So maybe that is part of what's happening here.

renrawnalon:
It looks like what is happening is, when you are defining your colors, you are using the sRGB color space (which might be the default for Sketch, I'm not sure though), but R.swift expects colors to be defined in the Generic RGB color space. This is more of an issue with the format colors are saved in than an issue of R.swift generating the wrong colors.

I was able to reproduce the conversion from `#F2F6F7` to `#EFF4F5` in Xcode by doing the following:

1. Open the Color Picker in Xcode (through storyboard or anything else).
2. Select the second tab and choose RBG Slider.
3. Click the settings button and select `sRGB`.
4. Enter `#F2F6F7` into the Hex Color field and press Enter.
    <img width="512" alt="srgb" src="https://cloud.githubusercontent.com/assets/605237/23327858/4967dbb0-fb57-11e6-928c-b74d80c80cca.png">
5. Click the settings button again and this time, select `Generic RGB`.
    <img width="513" alt="after changing to generic rgb" src="https://cloud.githubusercontent.com/assets/605237/23327860/49b7d69c-fb57-11e6-945a-a52951a06e47.png">

You should see that the Hex value changes to `#EFF4F5`. This conversion happens implicitly when you save to your .clr file, so R.swift ends up generating colors for the converted values.

[...]

PR #319, opened 2017-04-18
[WIP] Re-add support for all colorspaces

### LINKED ISSUE #317: Add support for P3 colors in CLR file
A check is now in place if colors are RGB, but we should also support P3 colors since those are also supported by iOS 10+.

Related to: https://github.com/mac-cain13/R.swift/issues/301 and #309

renrawnalon:
@mac-cain13 @tomlokhorst 

Upon further research into #301 and #309, I believe that the original implementation of colors that normalized all colors to the NSCalibratedRGBColorSpace (aka Generic RGB) was correct.

```
colors[key] = colorList.color(withKey: key)?.usingColorSpaceName(NSCalibratedRGBColorSpace)
```

What this is doing is it takes an `NSColor` in any colorspace, then converts it to Generic RGB. If the original color was already RGB, you get the same color as a result. If you look at the test cases I've added in this PR, you'll see that R.swift in fact creating the correct RGB values for `red`, `green` and `blue` in the Display P3 and sRGB `.clr` files.

When converting between colorspaces, depending on the color, there can be some loss in precision, which is reflected in the `blue` test case for the sRGB `.clr` file that I added. The conversion to RGB results in a blue value of `254.538`, which rounds up to `255`. Since this loss is minimal, I believe that it should be ignored. This should be fine because this should be the "real" value of the intended color, because it was created in a different colorspace. 

In this PR, I have returned the color conversion to its original form and added tests for the Display P3 `.clr` file, as well as a new sRGB `.clr` file and tests to go along with it. I've also updated the Colors.md documentation to indicate that when colors are created in a colorspace that isn't RGB, R.swift's generated RGB values will appear to be wrong, but are in fact correct.

I'm almost 100% sure that my logic regarding this issue is correct, but I would appreciate some independent verification. Do you guys think you could take a look and see if there's anything wrong with this PR, either logically or implementation-wise?

Thanks!


renrawnalon:
Hmm, so I don't think that this is correct. 

For one, this doesn't actually add support for Display P3 colors. In order to do that, I need to have R.swift generate the following code if the colorspace is Display P3.

```      
/// <span style='background-color: #FF0000; color: #00FFFF; padding: 1px 3px;'>#FF0000</span> Red
/// 
/// UIColor(displayP3Red: 1.0, green: 0.0, blue: 0.0, alpha: 1.0) or UIColor(red: 1.0, green: 0.0, 
static func red(_: Void = ()) -> UIKit.UIColor {
  if #available(iOS 10.0, *) {
    return UIKit.UIColor(displayP3Red: 1.0, green: 0.0, blue: 0.0, alpha: 1.0)
  } else {
    return UIKit.UIColor(red: 1.0, green: 0.0, blue: 0.0, alpha: 1.0)
  }
}
```

This leaves two issues:

1. What to do with sRGB? Convert it to RGB like before?
2. What to generate for comments in the case of Dislpay P3? Something like the above example?

Any thoughts?

tomlokhorst:
Is there a difference in colours or range between "RGB" and "sRGB", or are they the same?

For P3 colours, I think we should generate those with an `@available` annotation, so that they only appear on iOS 10.

Regarding the comments; I don't think the background-color css properties are working anymore in the current Xcode. If so, we can remove them.

renrawnalon:
Generic RGB and sRGB definitely seem to be different. sRGB seems to be somewhat of a standard because of its limited range, meaning that most other wide ranged colorspaces fully encompass it, so it's somewhat of a lowest common denominator. [Here](https://en.wikipedia.org/wiki/SRGB) is a Wiki article about sRGB. Generic RGB is described in [this Apple document](https://developer.apple.com/library/content/qa/qa1430/_index.html), although it doesn't give much information about how or when to use it in iOS apps.

Here's a quote from UIKit's [UIColor reference](https://developer.apple.com/reference/uikit/uicolor/1621925-init). 

> On applications linked for iOS 10 or later, the color is specified in an extended range sRGB color space. On earlier versions of iOS, the color is specified in a device RGB colorspace.

When combined with the information from [this blog post](http://inaka.net/blog/2014/09/05/getting-the-right-colors-in-your-ios-app/) about how to get the correct colors in your iOS app with Xcode from 2014 (prior to iOS10), it seems that Generic RGB is correct for iOS 9 and lower when using `UIColor(red:green:blue:alpha:)` and sRGB is appropriate for iOS 10 and above.

Here's a [pretty good article](https://developer.xamarin.com/guides/ios/platform_features/introduction-to-ios10/wide-color/) explaining the difference between Display P3 and sRGB.

I feel like what all of this means is:

- If the color space is Display P3 and the device is >= iOS10, then `UIColor(displayP3Red:green:blue:alpha:)` should be used without converting the colorspace.
- (I think that) If the color space is sRGB and the device is >= iOS10, then `UIColor(red:green:blue:alpha:)` should be used without converting the colorspace. (More investigation should be done for this case and other colorspaces on iOS10 before implementing.)
- Display P3 should be ignored for iOS9 and below.
- Other colorspaces should be converted to Generic RGB for iOS9 and below. (This should be tested more before implementing.)

I'll try playing around with various colorspaces to try and see what the best combination for different OS versions and colorspaces is.

Please let me know if you have any other thoughts on the subject.

tomlokhorst:
Your approach sounds good!

In general the goal of R.swift is the generate static references to existing resources.
If we can't be sure we're generating code that 100% of the time works, we'd rather not generate code at all. See [coding guidelines](https://github.com/mac-cain13/R.swift/blob/master/Documentation/Contribute.md#coding-guidelines).

Applying this to colours; Only generate code for colours if we're 100% it produces the _exact_ color that was in the resource file. E.g. If we're not sure if the color reproduction is exact on iOS 9, then don't generate code for iOS 9.

amarcadet:
What's the status of this issue?

I'm using R.swift 3.2.0 and I'm facing a similar issue.

I have created my .clr file using Apple Color Picker tool, my colors are all defined using the `sRGB` colorspace. However, if I set colors using Interface Builder or using R.color, I don't get the same color at all.

I have tried to look into some workaround and the only solution is to use a different .clr file for R.swift with `Generic RGB` colorspace. This is not an acceptable solution because I have to maintain 2 files (sRGB and Generic RGB).

SwiftGen offers the same feature but when I put the sRGB or Generic RGB file, I get the exact same output. I think this is somehow related to the colorspace conversion done [here](https://github.com/SwiftGen/SwiftGen/blob/d33ef5b8d995896176d2e6efe0ce08366a99ed6d/Pods/SwiftGenKit/Sources/Parsers/ColorsFileParser.swift#L151-L167).

mac-cain13:
During WWDC17 Apple announced support voor named Color Assets in Xcode asset folders. I think this is the way forward for predefined colors. So in R.swift 4 we will deprecate `clr` support and add support voor the named Color Assets. This has already been released in the 4.0.0.alpha.2 thanks to @tomlokhorst.

We could merge a fix that first converts to RGB color space like SwiftGen does, I'm okay with that, but on the long term we will drop the CLR support.

tomlokhorst:
Closing due to inactivity. 

Like @mac-cain13 said, the way forward is Color Assets in Xcode 9 / iOS 11. For iOS 10 support, you can use something like [ColorAssetCatalog](https://github.com/samrayner/ColorAssetCatalog). I haven't used that myself, it's just something I came across.

We can merge full colourspace support for `R.clr` if there's a complete, well tested implementation. But since R.clr is already deprecated for R.swift 4, I don't want to spent too much time on it.

PR #327, opened 2017-05-26
Regex fix for trailing zeros

guidosabatini:
With the actual regex, double parameters using precision with more than 1 digit per type cause the parameter not being considered in the R.generated.swift output

**Examples:**
%09.3 -> 9 total digits, 3 decimal digits and trailing zeros to get a length of 9
%12.10 -> 12 total digits with 10 decimal digits

Now these cases are considered.

input: `"example" = "string %@ float %012.10";`
**BEFORE:**
output: `static func example(_ value1: String) -> String`
**AFTER:**
output: `static func example(_ value1: String, _ value2: Double) -> String`


tomlokhorst:
Great fix! Thanks for the PR.

I've added tests, and will merge this.

PR #350, opened 2017-08-28
Update Examples.md

edulpn:
Documentation lacked further explanation on how to make R.swift identify the cell as reusable and generate need code to register and dequeue through NibResourceType protocol.


mac-cain13:
Thanks!

PR #353, opened 2017-09-03
Introduce 'generate' command

tomlokhorst:
Adds `generate` to rswift executable, to make space at the toplevel for other commands.

- [x] Add 'generate' command
- [x] Update readme
- [x] Add to R.swift 4 migration guide


Sources/rswift/main.swift
```diff
@@ -142,4 +142,33 @@ command(
...
+let exception = parser.hasOption("version") || parser.hasOption("help")
+
+if !exception && parser.shift() != "generate" {
+  var arguments = CommandLine.arguments
+  arguments.insert("generate", at: 1)
+  let command = arguments
+    .map { $0.contains(" ") ? "\"\($0)\"" : $0 }
+    .joined(separator: " ")
+
+  let message = "⚠️ R.swift 4 requires \"generate\" command as first argument to the executable.\n"
```
mac-cain13: Does Xcode pick this up as a build error? I think the build will fail due to the `exit(EXIT_FAILURE)`, but I'm not sure if the `fputs("\(message)\n", stderr)` will be shown as a build error.

In other places we print `error: message` to the console to get Xcode to show it.

tomlokhorst:
This is the error that's shown when the `generate` command hasn't been added tot the build script:

screen shot 2017-09-04 at 08 43 10

Issue #204, opened 2016-05-10
Add support for *.txt and *.json colors format


moto0000:
Hi,
plese add support for *.txt and *.json file formats with colors.

Something like that:
R.Colors.txt

```
Cyan-Color       : 0xff66ccff
ArticleTitle     : #33fe66
ArticleBody      : 339666
ArticleFootnote  : ff66ccff
Translucent      : ffffffcc
```

R.Colors.json

```
{
    "Cyan-Color"       : "0xff66ccff",
    "ArticleTitle"     : "#33fe66",
    "ArticleBody"      : "339666",
    "ArticleFootnote"  : "ff66ccff",
    "Translucent"      : "ffffffcc"
}
```

tomlokhorst:
I've seen this idea before, but I don't quite understand the reasons behind it.

Currently, R.swift only parses different file formats from Apple. It generates strongly typed references for identifier strings from files such as .xcodeproj, .storyboard, .xib.

The idea being: You'd rather have strongly typed references everywhere in your Swift code, but Apple doesn't support that in things like storyboard files. So, as a workaround, given that you need these stringly typed files anyway, lets take those as source to generate strongly typed equivalents.

Inventing our own stringly typed file format and parsing that seems a bit backwards.

My preferred method for having a consistent, shared list of colors is putting them in a Swift file, something like:

```swift
struct Colors {
  static let CyanColor = UIColor(hex: 0xff66ccff)
  static let ArticleTitle = UIColor(hex: 0x33fe66)
  static let ArticleBody = UIColor(hex: ox339666)
  static let ArticleFootnote = UIColor(hex: 0xff66ccff)
  static let Translucent = UIColor(hex: 0xffffffcc)
}
```

mac-cain13:
I agree with @tomlokhorst, but I'm open to hear about why a txt file or json file is better in your workflow to see if it would be something to support. Maybe you could tell us a little bit more about it @moto0000?

For reference you could also look into the [discussion about colors in the strings issue](https://github.com/mac-cain13/R.swift/issues/41#issuecomment-186824414) and the [CLR color issue](https://github.com/mac-cain13/R.swift/issues/169) where some of us already discussed some posibilities.

moto0000:
Hi,
In our team we working on Windows (Designer, Android Developer) and OS X devices, so designer often sends to us colors palette in txt files.

The txt files is easy to edit, now I'm cerate UIColor extension but when i use R.swift I would like to have everything in one place :)

[...]

PR #355, opened 2017-09-03
Move @available annotations for R.color

tomlokhorst:
As discussed in https://github.com/mac-cain13/R.swift/issues/204#issuecomment-326504069

### DIFF
--- a/Sources/RswiftCore/Generators/ColorStructGenerator.swift
+++ b/Sources/RswiftCore/Generators/ColorStructGenerator.swift
@@ -48,7 +48,7 @@ struct ColorStructGenerator: ExternalOnlyStructGenerator {
     }
 
     return Struct(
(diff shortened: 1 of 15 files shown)


Sources/RswiftCore/Generators/ColorStructGenerator.swift
```diff
@@ -48,7 +48,7 @@ struct ColorStructGenerator: ExternalOnlyStructGenerator {
     }
 
     return Struct(
-      availables: ["tvOS 11.0, *", "iOS 11.0, *"],
+      availables: [],
```
mac-cain13: Gebruiken we deze nu nog ergens anders? Anders kan ie er mss ook wel uit.

tomlokhorst: No, we're currently not using the available for structs anywhere. So they can be removed.

PR #370, opened 2017-11-29
update Examples.md

Hstripe:
Changed "Reusable table view cells"

When I use like before, it always calls UITableView's 'dequeueReusableCell' method.When I use like "Reuseable collection view cells", It worked.

So I commit this PR~


tomlokhorst:
Thank you for contributing to R.swift.

This is a good fix, I've merged into https://github.com/mac-cain13/R.swift/pull/375

Issue #40, opened 2015-08-03
Automate addition of run script phase


remzr7:
the xcodeproj gem should be able to do this using prepare_command

mac-cain13:
Nice idea! Especially since xcodeproj is already available if you use cocoapods!

bcapps:
This would be awesome to have. :+1:

PR #383, opened 2018-01-03
Automate R.swift installation in Xcode project

tomlokhorst:
Implements https://github.com/mac-cain13/R.swift/issues/40

- [x] Automatically add buildstep & file references to Xcode project
- [x] Generate empty placeholder file, explaining user needs to build
- [ ] Integrate with CocoaPods script_phase
- [ ] Test (with different real-world projects)
  
Example:
`rswift install MyProject.xcodeproj --target MyTarget`


mac-cain13:
I think we should revive this PR. Not aiming fully automatic installation through CocoaPods, but to help with the integration.

What I'm imagining:
1. Install R.swift using your favorite tool (easy as you probably use that tool quite often)
2. Add R.swift.Library using your favorite package manager (easy as you probably use that tool quite often)
3. Run `rswift integrate` and we'll ask you what workspace/project/target to integrate with and fix it for you (optionally you can pass arguments instead of answering questions)

Think this makes integration way easier for small projects and first time users.

tomlokhorst:
Perhaps even one of those scary internet installation scripts, like [brew](https://brew.sh/) has?

So that you only need one step.

`rswift install` can also add the R.swift.Library dependency using SPM or CocoaPods

tomlokhorst:
No longer relevant

PR #385, opened 2018-01-20
Update documents which has unnecessary optional collectionView.

ktanaka117:
`collectionViews` which are written in `Example.md` are not need to be Optional. 

https://github.com/mac-cain13/R.swift/blob/master/Documentation/Examples.md


tomlokhorst:
Hi, thanks for your contribution!

In the example, the RecentsController is a subclass of UICollectionViewController which has an [optional collectionView property](https://developer.apple.com/documentation/uikit/uicollectionviewcontroller/1623983-collectionview). So I think the example is correct as it is now.

ktanaka117:
Oh..., sorry it's my misread. x<
I always use UICollectionView instead of UICollectionViewController.
I close this issue. 

Thank you for always useful library.

Issue #295, opened 2016-12-13
Support for Cocoapods resource_bundles


wangguangfeng:
I use R.swift in a dynamic framework which organizes all the images for the main project and other frameworks. After the dynamic framework building successfully, but the image I access in main project is nil.
This is the code: `let image = R.image.driver_task_calender()`
After check, I found the` let hostingBundle = Bundle(identifier: "com.**.**") ?? Bundle.main` is wrong.

I think this is the correct hosting bundle ：

```
var hostingBundle: Bundle {
        let podBundle = Bundle(for: TheClassName.self)
        if let bundleURL = podBundle.url(forResource: "TheBundleName", withExtension: "bundle") {
            if let bundle = Bundle(url: bundleURL) {
                return bundle
            }
        }
        return podBundle
    }
```

Expect to get help.

wangguangfeng:
I organized the framework by cocoapods, in the dynamic framework place a spec file. In the spec file:
s.resource_bundles = { 'JFoundation' => ['JFoundation/Resources/**/*.{storyboard,xib,xcassets,json,imageset,png}'] }

mac-cain13:
Thanks for using R.swift and reporting this issue.

I've fixed this issue in the latest release of R.swift, version 3.2.0. If you upgrade to that version your problem should be solved! Please let me know if you still have troubles using R.swift from a framework after upgrading. :)

wangguangfeng:
please do not close the issue. After upgrading there are something wrong, I think you should upgrade the R.swift and add some script, like adding --accessLevel public to the call to R.swift to use R.swift in a library. To distinguish the different situation between using in main project and dynamic framework.

the correct hosting bundle : 
```
var hostingBundle: Bundle {
        let podBundle = Bundle(for: TheClassName.self)
        if let bundleURL = podBundle.url(forResource: "TheBundleName", withExtension: "bundle") {
            if let bundle = Bundle(url: bundleURL) {
                return bundle
            }
        }
        return podBundle
    }
```

your hostingBundle generated by R.swift : 
`fileprivate static let hostingBundle = Bundle(for: R.Class.self)`

wangguangfeng:
Expect to get help.

mac-cain13:
Ah alright, so basically you're asking for support for Cocoapods `resource_bundles` feature.

[...]

Issue #186, opened 2016-04-04
Add support for bundle folder


vladimirvrabel:
Hi, 

please add support for items in Resources.bundle e.g.: Resources.bundle/images/myimage@2x.png. Because when we use Resources.bundle, then there are not changes at pbxproj file for each change in this directory with using git

Thanks

mac-cain13:
I'm not sure how many people would benefit from supporting bundles, if you have a pure Swift project you can just use frameworks and generate a `R.generated.swift` file within the project exposing the resources in there. A seperate bundle with resources sounds to me like a thing from the ObjC world with static linked frameworks that couldn't include resources.

Maybe I'm wrong, love to hear about the use cases, but I don't think this is a top priority.

martinflorek:
We use bundle folders for easy collaboration with designers. They behave like referenced folders in Xcode, no change to the .pbxproj file for easy GIT. No need to handle .xcassets JSON structure, no slow drag-&-drop of resources in Xcode, no problems with resources renaming and moving. Designers just export PNGs into the bundle and push to GIT repository. No need to learn designers anything about the Xcode and GIT conflicts resolving.

Bundles have too many pluses to not to use them.

kristiyanP:
@martinflorek 
If your designers use sketch you may want to check these two plugins:
https://github.com/iGenius-Srl/Seca
https://github.com/iGenius-Srl/sketch-app-asset-export

**The resources exported with these plugins generate also the json structure ( for ios only) and take care of the naming conventions**

### What we do:
We have a git repo where designers export using these plugins
Then for both iOS and Android we have a python script that just copies the files in the project at build time.

mac-cain13:
Together with @tomlokhorst I looked into bundle support. Also relates to CocoaPods w/bundles as discussed in #295.

We think R.swift should detect the bundle by itself, since bundles are also mentioned in the Copy Resources build phase in Xcode. So we can detect them automatically and then put all resources found in that bundle under `R.bundle.myBundle.*`.

This more robust than the proposed solution in PR #395 since this won't support having both resources in your project as well as in one or more bundles. Detecting bundles, picking out all resources and adding them in the R file should not be too hard to implement.

cooler333:
What about static frameworks? It's not contain resources

cooler333:
[...]

PR #395, opened 2018-03-18
Support ResourceBundles

svendr:
This PR tries to add support for Resource Bundles.

Organizing project resources with Resource Bundles is a fairly common approach in iOS development, especially if you are developing a custom Framework (using Resource Bundles here simplifies integration into client projects heavily). Cocoapods also [encourages using Resource Bundles](https://guides.cocoapods.org/syntax/podspec.html#resources).

R.swift doesn't have native support for loading resources from a Resource Bundle embedded inside a Framework or Executable Target Bundle (Main Bundle). This is a known issue, see #295. I've seen several workarounds manipulating the hostingBundle-definition with sed etc.

So I've introduced the new CLI-parameter **--resourceBundleName**. When given, it changes the hostingBundle from `Bundle(for: R.Class.self)` to `Bundle(url: Bundle(for: R.Class.self).url(forResource: "ResourceBundleName", withExtension: "bundle")!) ?? Bundle(for: R.Class.self)` (replace "ResourceBundleName" with the value of the --resourceBundleName option).

I've successfully tested this with several projects I'm working on.

Glad to hear your thoughts / feedback on this.


moshiwu:
Great PR, which solve my problem exactly！

tomlokhorst:
Thanks for this PR, it looks great!

I'll test this with some of our projects as well.

mac-cain13:
Nice! Besides the testing @tomlokhorst will perform I think it would be nice to have this documented somewhere and maybe it's possible to create a testcase for this in the ResourceApp we have? If we use it somewhere it's less likely to break.

svendr:
Thanks for your feedback! I'll try to add a test and extend the doc

tomlokhorst:
You can ignore the Bitrise CI error about the failing tests.
The tests aren't properly setup for Xcode 9.

Btw, I tried to test this, but couldn't get it to work...
So I would appreciate some documentation! 😁

svendr:
Hey Tom, I've extended the ResourceApp with 2 new targets

- The `ResourceBundle` target contains some dummy resources
- `ResourceBundleApp` references the `ResourceBundle` build product as target dependency
- `ResourceBundleApp` also calls rswift with the new `--resourceBundleName` option. Important thing here is to also use the `--target` option.

I've also set up a simple unit test and reworked the Podfile to silence several annoying warnings.

I'm still stuck on a problem with `R.validate()`, because it's internals sometimes rely on implicitly referencing the MainBundle. E.g. `StoryboardGenerator.swift` produces code like:

```
static func validate() throws {
        if UIKit.UIImage(named: "User") == nil { throw Rswift.ValidationError(description: "[R.swift] Image named 'User' is used in storyboard 'Main', but couldn't be loaded.") }
      }
```

But I can't just exchange it with `UIImage(named: "User", bundle: R.hostingBundle)` because the R-struct doesn't exist when building the rswift-Binary. I've also tried to workaround it with Reflection, but had no success. 

I think we would have to extend the R.swift.Library and introduce some kind of `BundleHelper`-struct, which could be accessed from R.generated.swift and be accessed from `StoryboardGenerator.swift` to set / get the `hostingBundle`?

Or do you see a simpler option?

tomlokhorst:
I couldn't quite get this to work in my own CocoaPod with embedded resources.
Could you describe how this is supposed to work?

Does the pod project need a Xcode project with a R.swift build step? Or is this part of the R.swift build step of the application?

magrinj:
@volodg @ANSCoder Hello guys ! What about merging this PR ? :)

andlang:
Would be great to see that feature in the upcoming 5.1.0 release

letko-dmitry:
@svendr any news?

yuhanle:
Good pr, it solve my question.

mac-cain13:
We've looked into this, thanks for all input and discussion around this topic. We think bundles should be properly detected by R.swift and be nested in the R-struct. The design of this will be discussed in https://github.com/mac-cain13/R.swift/issues/186.

I'll close this PR because it will make it hard to use resources in the app itself combined with resources in one or more bundles. Since you then will have 2 structs named `R` in your project.

Issue #394, opened 2018-03-15
Avoid Empty Structs


fabb:
It would be nice for autocompletion, if R.generated.swift would not contain any empty structs.

E.g. if we don't have any xibs or storyboards, this will be generated:
```
/// This `R.nib` struct is generated, and contains static references to 0 nibs.
  struct nib {
    fileprivate init() {}
  }
```

Typing `R.` in Xcode will still suggest autocompletion `.nib`, even though it is empty. So it would be nice if the `struct` wouldn't be generated at all if empty.

fabb:
PR was merged

PR #396, opened 2018-03-21
Attempt to not include empty structs into the generated file.

soberman:
HI! I am trying to alter the generation of the structs so that the generated R file wouldn't include the empty ones. Strangely enough, when I reference this commit in the `Podfile`, so that I would have this version instead of the `4.0.0` tag, after `pod install` I am not able to build anymore. The original project after my changes compiles just fine, it is only through Cocoapods - `rswift` file is simply not there.
Could you help me out please? Maybe I have missed something. Really struggling here. Thanks!

### DIFF
--- a/Sources/RswiftCore/Generators/AggregatedStructGenerator.swift
+++ b/Sources/RswiftCore/Generators/AggregatedStructGenerator.swift
@@ -23,12 +23,12 @@ class AggregatedStructGenerator: StructGenerator {
 
     let collectedResult = subgenerators
       .flatMap {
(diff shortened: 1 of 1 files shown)


fabb:
resolves #394

tomlokhorst:
Unfortunately, you can't test a custom build using CocoaPods. 
Because R.swift is an executable, not a library, we ship it as a zip file that is downloaded through CocoaPods, but this only works for released versions.

The best way to test your changes to `rswift` is to manually create a build, and change the build step in your test app to point to the new `rswift` executable.

tomlokhorst:
I've tested this, and it works great!

Thanks for your contributing to R.swift!

PR #401, opened 2018-04-05
 Added support for generating SwinjectStoryboard storyboard initializers 

lgrzywac:
There is new generator parameter:
- storyboardInstantiationAdditions

Possible values:
- noregular - do not generate regular initializers
- swinject - add swinject initializers with WithSwinject suffix. Ignored when combined with "replaceRegularWithSwinject"
- replaceRegularWithSwinject - replaces regular initializer with swinject initializer

### DIFF
--- a/ResourceApp/ResourceAppTests-tvOS/ResourceAppTests_tvOS.swift
+++ b/ResourceApp/ResourceAppTests-tvOS/ResourceAppTests_tvOS.swift
@@ -26,6 +26,10 @@ class ResourceAppTests_tvOS: XCTestCase {
       for warning in expectedWarnings {
         XCTAssertTrue(logLines.contains(warning), "Warning is not logged: '\(warning)'")
       }
+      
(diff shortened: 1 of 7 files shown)


lgrzywac:
@mac-cain13 , can you provide details about failed checks in CI?

mac-cain13:
The CI is always complaining about PRs not in this repository, so don't worry about that.

I've looked at the PR, I think it doesn't really fit in the scope of the R.swift. This generates code for a specific library, we want to stay close to what Apple offers and offer a types version of their APIs. I would like to stay clear of supporting specific libraries by generating code for them.

It would be interesting to see if it is possible to use the types offered by the [R.swift.Library](https://github.com/mac-cain13/R.swift.Library) to create an handy extension that will help you with using Swiftinject. This is also the route that [SegueManager](https://github.com/tomlokhorst/SegueManager) for example uses.

PR #418, opened 2018-06-12
Add explicit file inclusion to rswiftignore

renrawnalon:
Hi. This pull request adds `!` functionality for `.rswiftignore` similar to that of `.gitignore`, but slightly more simplistic. The main use case for this functionality that I have is for generally excluding all storyboard and xib files that are associated with Objective-C classes while selectively including some storyboard and xib files that should be accessible from Swift. I am currently in the process of converting a large code base from Objective-C to Swift and it is much cleaner to write something like this:

```
# .rswiftignore

**/*.storyboard
!IncludeThisSwift.storyboard
```

than it is to import every UIViewController subclass in the app into the `Project-Bridging-Header.h` file, even though the majority of those files won't ever be accessed from Swift.

I have updated the documentation and the ResourceApp to reflect the changes that I have made, but I couldn't find anywhere that unit tests the `.rswiftignore` functionality, so I haven't implemented any unit tests myself. I'm not quite sure how I would go about unit testing this behavior, but if you would like me to, I would be happy to try and come up with something.

Please let me know if you think this will be a reasonable addition to the project, and if so, let me know if you'd like me to change anything about how I've implemented this feature.

Thanks!


renrawnalon:
Implements https://github.com/mac-cain13/R.swift/issues/420

mac-cain13:
Thanks for the PR! I'll look into it to see if we can merge it into the 5.0 release.

mac-cain13:
Thanks, looks good! I've merged it into master, so it will be in the next release.

Thanks again for contributing.

PR #424, opened 2018-07-04
Fixes parsing problem in NibParserDelegate.

rafaelnobrepd:
Adds a test case with the kind of problem encountered when running R.swift 5.0.0-alpha1 in Xcode 10 and a attempt to fix it.


tomlokhorst:
This is a great PR!

Thanks for identifying the issue and creating this fix.

PR #431, opened 2018-07-25
Docs - QA - improve: use in library

teameh:
- State you ideally need a dedicated xcode project to include build script

- Add link to issue explaining this

- Change example script with `accessLevel` for Cocoapod users: 
If you have a dedicated xcode project for your library and you use R.swift in your library you need to install it somehow right? Probably with a regular Podfile? and thus the script needs to be changed to `"$SRCROOT/Pods/R.swift/rswift" generate "$SRCROOT"`


tomlokhorst:
Thanks for this improvement!

PR #451, opened 2018-09-23
Performance Tweaks

bclymer:
# What

This PR brings 2 performance enhancements.

## Ignored Files

When ignoring files using `.rswiftignore`, all potential ignored files are calculated during the creation of `IgnoreFile`. It iterates recursively through all subdirectories in `.rswiftignore`'s root directory. The problem here is that will often include `Pods`, and depending on the project could include `node_modules` as well. These directories can of course be massively complex compared to the actual user project.

Instead of iterating recursively through all subdirectories from root, the new solution is to do a shallow iteration through the root directory, ignore `Pods` and `node_modules`, and then iterate recursively through the rest.

## Identifying Resources

Large projects can have a significant amount of resources. In `Resources.init`, all the resources were being iterated over 6 times. Iterating over them once brings a marginal speedup.

# Benchmarks

### Before
screen shot 2018-09-23 at 1 18 59 pm

### After
screen shot 2018-09-23 at 1 21 05 pm

### Before
screen shot 2018-09-23 at 1 19 22 pm

### After
screen shot 2018-09-23 at 1 20 26 pm

These benchmarks are using the project I spend the majority of my time in, which is rather large. However, the speed ups should apply universally to all projects. Though projects using a `.rswiftignore` file that aren't using Cocoapods probably won't get much benefit.

### DIFF
--- a/Sources/RswiftCore/ResourceTypes/Resources.swift
+++ b/Sources/RswiftCore/ResourceTypes/Resources.swift
@@ -26,15 +26,43 @@ struct Resources {
   let reusables: [Reusable]
 
   init(resourceURLs: [URL], fileManager: FileManager) {
-    assetFolders = resourceURLs.compactMap { url in tryResourceParsing { try AssetFolder(url: url, fileManager: fileManager) } }
-    images = resourceURLs.compactMap { url in tryResourceParsing { try Image(url: url) } }
-    fonts = resourceURLs.compactMap { url in tryResourceParsing { try Font(url: url) } }
-    nibs = resourceURLs.compactMap { url in tryResourceParsing { try Nib(url: url) } }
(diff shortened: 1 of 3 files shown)


tomlokhorst:
Thanks for your PR, the screenshots look like a nice performance improvement!

I noticed a lot of the GlobTests now fail with your changes to the `expandGlobstar` function. Could you fix those? And perhaps move the blacklist to a parameter of that function, instead of hardcoding it.

bclymer:
@tomlokhorst fixed the broken tests (surprise, the tests were working and my code wasn't) and added a new test to verify the blacklist is working as expected. 

Also made the blacklisted directories an argument as you requested, but added it to Glob's `init` instead of the function itself to make testing possible. It provides the default value of `["node_modules", "Pods"]`.

PR #452, opened 2018-09-23
Sort results so output is consistent every run.

bclymer:
# What

I noticed on the latest alpha, `R.generated.swift` was slightly different every time. As part of my build, I actually run a local diff on the newly generated `R.generated.swift` and the existing one, and only use the new one if something has changed. Otherwise, Xcode would always see that a file had been touched, and the incremental builds would not be as efficient as they could be.

This PR just sorts things that were mapped to a non-ordered data type at some point.

I might have missed some, but this was at least giving my project consistent results.


bclymer:
I see this is now actually built in - https://github.com/mac-cain13/R.swift/blob/master/Sources/RswiftCore/RswiftCore.swift#L64

Which is good news, because I can remove that check within my built script (I adopted R.swift before that check), however that means this PR actually fixes what would be considered a "bug" in the alpha. Since that feature is no longer working as expected.

tomlokhorst:
Thank you for your pull request!

This was indeed an accidental error. Which is no fixed for the 5.0 release (whenever that will be).

PR #468, opened 2018-12-10
Xcode10 support

### LINKED ISSUE #456: ⚠️ R.swift and Xcode 10 / Swift 4.2
The currently latest alpha release of R.swift (5.0.0.alpha.2) supports Swift 4.2 and Xcode 10.

However, there is a known build system issue. We want to fix this issue _before_ releasing the official 5.0 version of R.swift.

If you can work around the build system issue, feel free to use the alpha version of 5.0, it works pretty much like R.swift 4.0, but for Swift 4.2. (See [complete list of differences](https://github.com/mac-cain13/R.swift/releases))

# Issue: R.swift build step and the New Build System
When the new build system is used, with a standard R.swift build step, this build step is not run _before_ the Compile Sources build step.

Because of this, the initial compilation will fail, because there's no R.generated.swift file, and references to the `R` struct in the rest of the source code are invalid. 

While it's easy to "build twice" during development in Xcode, this is particularly a problem on CI servers, where the build will simply fail because of the missing R.generated.swift file.

## Partial solution: Adding R.generated.swift as an output file
As a fix for the above issue, you can add R.generated.swift as an output file:

This will make sure the R.swift step has finished, before the Compile Sources step starts.
However, with this output file, the step will only run once:

## Issue: Build step with output file only runs once
When R.generated.swift is configured as the output file (and there are no input files), the R.swift build step will only run once. On subsequent incremental builds, the step is skipped. Because of this, new resources won't show up in the R struct.

It is sometimes possible to list the resource files used als input files for the build step, but not always.

For example, when adding a new image to the asset catalog, the new filename isn't known beforehand.

# Workarounds
These are possible workarounds:

1. Use the Legacy Build System. This behaves like before, and the R.swift build step will always run.
2. Commit the R.generated.swift file. This means during development, you can "build twice" and commit the resulting up-to-date file to git.
3. Clean build when resources change. Configure the output file for the build step, and manually do a clean build whenever one of the resources are changed.

None of these workarounds are ideal, so we're still looking for a proper solution.

mac-cain13:
Fixes #456 by making it possible to define an input file (touched by R.swift on every run) and output file (the generated file) for the script build phase. This solves the Xcode 10 issues.

R.swift will fail with an error if the input or output file are not set for the build phase.

Todo:
- [x] Check if we want to cleanup the code
- [x] Check if errors are clear enough
- [x] Make a note that we want to maybe generate real input file lists or was there an issue with that?! Not sure anymore...
 - [x] Write documentation on how to upgrade from R.swift 4 to R.swift 5


tomlokhorst:
Re: checklist item 3 

Our last status with respect to generated input file lists:

 - Use project.pbxproj as a dependency for newly added file resources (.storyboard, .xib etc.)
 - Use Asset catalog Content.json files in directories for newly added resources in folders of Asset catalogs

Initial investigation seems to indicate that some previously existing Content.json file will always be touched when a new file or folder is added to an Asset catalog.

Sources/rswift/main.swift
```diff
@@ -141,30 +145,18 @@ let generate = command(
...
-    + "Change your call to something similar to this:\n\n"
-    + "\(command)"
-    + "\n"
-
-  fputs("\(message)\n", stderr)
-  exit(EXIT_FAILURE)
+// Touch last run file
+do {
+  let tempDirPath = try ProcessInfo().environmentVariable(name: EnvironmentKeys.tempDir)
+  let lastRunFile = URL(fileURLWithPath: tempDirPath).appendingPathComponent("rswift-lastrun")
```
tomlokhorst: Make `rswift-lastrun` a constant somewhere.

Sources/RswiftCore/RswiftCore.swift
```diff
@@ -13,6 +13,17 @@ import XcodeEdit
 public struct RswiftCore {
 
   static public func run(_ callInformation: CallInformation) throws {
+    let lastRunPath = callInformation.tempDir.appendingPathComponent("rswift-lastrun").path
+    guard callInformation.scriptInputFiles.contains(lastRunPath) else {
```
tomlokhorst: This should handle both `$(TEMP_DIR)/rswift-lastrun` and `$TEMP_DIR/rswift-lastrun`

mac-cain13: This is already working, Xcode expands all variables for us, so we only see the resolved paths and not how the variable is noted.

PR #485, opened 2019-01-29
Add mint support

lilpit:
If you don't use Cocoapods in your project, there is no proper way to manage the R.swift dependancy.

A dependency manager based on SwifPM called [Mint](https://github.com/yonaskolb/Mint) is ganning traction in the community and is already supported by some other swift based tools like Carthage or SwiftGen.

This PR just adds a product to Package.swift, the only required thing needed by Mint to work. 
To make R.swift installable via mint, a new Release of R.Swift will be needed

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,6 +3,9 @@ import PackageDescription
 
 let package = Package(
   name: "rswift",
+  products: [
+    .executable(name: "rswift", targets: ["rswift"]),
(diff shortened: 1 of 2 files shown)


tomlokhorst:
This commit was merged as part of https://github.com/mac-cain13/R.swift/pull/538

Thanks for your PR!

PR #495, opened 2019-03-02
Remove warning on subclasses of unwind segue

miyakeryo:
When I use subclasses with unwind segue, I'm getting warning. But I can't select custom. So I removed warning.

image


tomlokhorst:
Thanks for your PR!

PR #514, opened 2019-04-22
'indictator' -> 'indicator' spelling fix

imryan:


tomlokhorst:
Ha! If only we had R.swift for the REAME, this error wouldn't have happened. 😀

PR #516, opened 2019-04-24
Objective-C Support

bclymer:
## Pre-Read

This PR isn't 100% ready to be merged. For example, `ENV_VARS` still exists, the `.podspec` still points to my fork, etc. This PR is just going up in order to:

1) Promote discussion about the possibility of Objective-C support for the main R.swift repo.
2) Welcome normal code review feedback.

If the concept is accepted, I will get another branch that excludes some of the unnecessary changes up and ready.

## What Is This?

This PR adds support for generating Objective-C compatible code that directly references the existing generated Swift code. For example, in Swift you can access an image like so:

```
R.image.conflicting()
```
Now, in Objective-C you can call
```
[RObjc image_conflicting];
```
and you will get the same `UIImage?` back.

You can also call
```
[RObjc image_conflictingWithCompatibleWith:traitCollection];
```
Two Objective-C compatible functions get generated for most Swift functions. The first with the minimal amount of parameters, and the other with all parameters. Since Objective-C doesn't have default values, this cleans up usage significantly. In the case where the minimum number of parameters is equal to all parameters (such as strings with required formatting inputs), only one function is generated.

The same type of support exists for all types that R.swift supports with the exception of `TypedStoryboardSegueInfo` for obvious incompatibility reasons.

## How Do I Use It?

Generating Objective-C compatible code is possible by appending `--objc` to your command line arguments. This means the generate is **opt-in** and will not affect most projects.

As stated above, most existing `R.x.y` calls have been ported to `[RObjc x_y]`. It should be as simple as adding `#import Project-Swift.h` to the Objective-C file you want to utilize R.swift in.

In addition, the generated header contains `@available(swift, obsoleted: 1.0, message: "Use R. instead")`, preventing Swift usage of the Objective-C compatible code.

### TypedStoryboardSegueInfo

As mentioned above, this type of structure isn't compatible with Objective-C, so these are excluded.

### Invalid File Names

Files with invalid names are not generated. For example the following will not be ported.
```
static func `__FILE__`(_: Void = ()) -> Foundation.URL?
```

### Validate Functions

Validate functions are not ported. I really just didn't see the point in doing it. I think it would be trivial to add back.

### Deprecated Functions

They're deprecated anyway, so why port them?

## Example R.generated.swift

This is only the Objective-C compatible part, it is created from the example app provided in the repo.

https://gist.github.com/bclymer/fa19c0b7260cc561b8dfce2b0e26385a

### DIFF
--- a/ENV_VARS
+++ b/ENV_VARS
@@ -1,3 +1,7 @@
+Arguments:
+generate "../../ResourceApp/R.generated.swift" --objc
+
+Env Vars:
 PRODUCT_BUNDLE_IDENTIFIER=nl.mathijskadijk.ResourceApp
 SCRIPT_INPUT_FILE_COUNT=1
 SCRIPT_INPUT_FILE_0=./temp/rswift-lastrun
--- a/ENV_VARS
+++ b/ENV_VARS
@@ -1,0 +1,14 @@
+PRODUCT_BUNDLE_IDENTIFIER=nl.mathijskadijk.ResourceApp
+SCRIPT_INPUT_FILE_COUNT=1
+SCRIPT_INPUT_FILE_0=./temp/rswift-lastrun
+SCRIPT_OUTPUT_FILE_COUNT=1
+SCRIPT_OUTPUT_FILE_0=../../ResourceApp/R.generated.swift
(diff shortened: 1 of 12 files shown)


bclymer:
Well I forgot that any merges I made into my fork would just appear in this PR. Going to close since there was no activity in 22 days anyway.

PR #528, opened 2019-06-13
Make R.Swift compatible with Swift 5.0

elenipapanik:
When trying to compile the R.generated swift file code with Swift 5.0 I stumbled upon the following error:

Screenshot 2019-06-13 at 13 09 23

Therefore, I changed the `NibStructGenerator` to generate `NSString `as key dictionary of the `optionsOrNi`l parameter instead of `UINIb.OptionsKey` type which fixes the compile error.


tomlokhorst:
Are you using version 5 of `R.swift.Library` together with R.swift 5?
The generated code should already be compatible with the library.

elenipapanik:
Yes, I am using R.swift.Library v5.0.1 and I am still getting the same error.

PR #535, opened 2019-07-20
Reorder Module Imports in Generated Source Code

### LINKED ISSUE #534: Imported Namespaces Should Prefix Rswift Import
## The Problem

I am using R.swift in a multi-framework application. For example, I have the following projects in my workspace:

* **MyApp**: the main iOS application with the AppDelegate, etc.
* **Core**: a dynamic framework with shared code and external libraries. The external libraries are Cocoapod dependencies that are being linked as static frameworks using the [cocoapods-static-swift-framework](https://github.com/leavez/cocoapods-static-swift-framework) plugin. This framework is used to share code between the main app and extensions.
* **Views**: a dynamic framework with my custom UIViewController and UIView classes. I have them in a framework so that I can play with them in a Playground and also to isolate untestable UI code for code coverage reporting.
* **Services**: code for calling external services. It's isolated into its own framework to be used in a Playground and for code coverage reasons.
* **Features**: my main business logic code.

My intent is that I want to link R.swift.library into my **Core** framework and then reference that in the `R.generated.swift` files that I generate for each of my frameworks, extensions, and the main application. And this works using the `--import` command line argument. The problem is that `--import` appends the `import Core` statement after the `import RSwift` line, which causes a compilation error. The `import Core` statement must be before `import RSwift` in order for the compiler to be able to find the `RSwift` module inside of the `Core` framework.

If I move the `pod 'Rswift'` line in my `Podfile` outside of the `Core` framework target, then `R.swift.library` gets added to my application and all of my frameworks and I receive conflict warnings at runtime because the linker finds multiple copies of `R.swift.library`.

## The Proposal

My proposal is that the `swift` tool be changed to add imports specified using the `--import` argument before the `import Rswift` statement.

mfcollins3:
I updated `ImportPrinter` to change the order of the imports that are
generated in the output source code. Previously all imports were being
sorted alphabetically and then output to the generated source code. This
was a problem if `R.swift.library` is being statically linked in another
dynamic framework and the containing framework's module name is
alphabetically after the `Rswift` module name. In this case, the Swift
compiler does not correctly import the `Rswift` module.

I changed `ImportPrinter` to always output modules that are specified
using the `--import` command line argument before modules that are
extracted from the source code. This should fix the import issue.

I updated `CallInformation.swift` to pass the list of modules as an array
instead of a set in order to keep the order in which they were specified
on the command line in order to ensure proper ordering of imports in
the generated source code.

Resolves #534

### DIFF
--- a/Sources/RswiftCore/CallInformation.swift
+++ b/Sources/RswiftCore/CallInformation.swift
@@ -16,7 +16,7 @@ public struct CallInformation {
   let rswiftIgnoreURL: URL
 
   let accessLevel: AccessLevel
-  let imports: Set<Module>
+  let imports: [Module]
 
   let xcodeprojURL: URL
   let targetName: String
@@ -39,7 +39,7 @@ public struct CallInformation {
     rswiftIgnoreURL: URL,
 
     accessLevel: AccessLevel,
-    imports: Set<Module>,
+    imports: [Module],
 
     xcodeprojURL: URL,
     targetName: String,
--- a/Sources/RswiftCore/SwiftTypes/CodeGenerators/ImportPrinter.swift
+++ b/Sources/RswiftCore/SwiftTypes/CodeGenerators/ImportPrinter.swift
@@ -13,19 +13,22 @@ import Foundation
 struct ImportPrinter: SwiftCodeConverible {
   let swiftCode: String
 
-  init(modules: Set<Module>, extractFrom structs: [Struct?], exclude excludedModules: Set<Module>) {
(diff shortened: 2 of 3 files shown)


mac-cain13:
Thanks for the PR seems like a fine proposal to fix your issue. I need some time to review the code & try it out. I would like to try to have this in the 5.1 release.

tomlokhorst:
Thanks for you detailed issue and PR!

I've added a comment in the code referring to this issue, for future reference.

PR #538, opened 2019-07-28
Add Mint Support (Includes Updated Documentation)

robfeldmann:
This PR includes the previous pull request from @lilpit and adds the following modifications:

- provides the additional documentation requested by @mac-cain13 in the original PR review
- adds .swiftpm to the .gitignore file (pretty sure this should be ignored)

As a refresher, @lilpit submitted #485 :

> If you don't use Cocoapods in your project, there is no proper way to manage the R.swift dependancy.
> 
> A dependency manager based on SwifPM called Mint is ganning traction in the community and is already supported by some other swift based tools like Carthage or SwiftGen.
> 
> This PR just adds a product to Package.swift, the only required thing needed by Mint to work.
> To make R.swift installable via mint, a new Release of R.Swift will be needed

**NOTE**: My `R.Swift.Library` [Pull Request 34](https://github.com/mac-cain13/R.swift.Library/pull/34) should also be accepted (probably first) because it is required for the instructions in part 2 here.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ rswift.xcodeproj
 .build
 build
 DerivedData
+.swiftpm
--- a/Package.swift
+++ b/Package.swift
@@ -6,6 +6,9 @@ let package = Package(
   platforms: [
     .macOS(.v10_11)
   ],
+  products: [
+    .executable(name: "rswift", targets: ["rswift"])
+  ],
   dependencies: [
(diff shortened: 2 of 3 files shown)


tomlokhorst:
Thanks a lot for this detailed PR!

Library installation instructions don't work yet because there isn't a tag for an SPM release. But that will be fixed before the R.swift 5.1 release.

Issue #219, opened 2016-05-20
Localised Strings in addition to currentLocale


uny:
I would like to use localized strings in addition to `NSLocale.currentLocale()`.

**Pattern 1:**

``` swift
struct string {
  struct localizable {
    let locale: NSLocale
    init(locale: NSLocale = NSLocale.currentLocale()) {
      self.locale = locale
    }
    /// Locales: en, ja
    static func helloWorld() -> String {
        return String(format: NSLocalizedString("HelloWorld", comment: ""), locale: self.locale)
    }
  }
}
```

**Pattern 2:**

``` swift
struct string {
  struct en {
    /// Locales: en, ja
    static func helloWorld() -> String {
        return String(format: NSLocalizedString("HelloWorld", comment: ""), locale: NSLocale(localeIdentifier: "en"))
    }
  }
}
```

tomlokhorst:
Hey, thanks for creating this issue!

I do plan on adding a feature to R.string for requesting a localized string in a specific locale. Currently, it always chooses the preferred localization.

For now, it's possible to use the generated `StringResource` to get the strongly typed key for a localized string. You can do something like this:

```swift
let basePath = NSBundle.mainBundle().pathForResource("Base", ofType: "lproj")!
let nlPath = NSBundle.mainBundle().pathForResource("nl", ofType: "lproj")!
let baseBundle = NSBundle(path: basePath)!
let nlBundle = NSBundle(path: nlPath)!

let baseTripsTitle = baseBundle.localizedStringForKey(R.string.planner.tripsTitle.key, value: nil, table: R.string.planner.tripsTitle.tableName)
let dutchTripsTitle = nlBundle.localizedStringForKey(R.string.planner.tripsTitle.key, value: nil, table: R.string.planner.tripsTitle.tableName)
```

uny:
Thank you!
Currently, I am doing the same as your suggestion.

I need time to read and understand the code, but I would like to contribute when I can.
Do you have any idea to implement this feature (pattern 1, pattern 2, or others)?

mbesnili:
Any progress on this issue ? Maybe you can give some ideas about how this feature should be implemented, so that someone else can be implemented.

tomlokhorst:
I haven't taken the time to think about this more, since I don't need this at the moment.

The project I'm currently working on does need localised strings for a specific locale, but it's still a few months before we'll start working on that feature.

[...]

PR #553, opened 2019-08-31
Localised Strings in addition to Locale.preferredLanguages

tomlokhorst:
Adds optional `preferredLanguages` parameter to generated strings functions. Used as a replacement for `Locale.preferredLanguages`.

Usage:

```swift
// Assuming the app is localised in English and French
// Returns French localisation, even if Phone is set to English
let hello = R.string.welcome.hello(preferredLanguages: ["nl", "fr"])
```

Fixes https://github.com/mac-cain13/R.swift/issues/219

### DIFF
--- a/LocalizedStringApp/LocalizedStringApp.xcodeproj/project.pbxproj
+++ b/LocalizedStringApp/LocalizedStringApp.xcodeproj/project.pbxproj
@@ -1,0 +1,677 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 50;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		4BBE4418B51EC4EA7E8ED322 /* libPods-LocalizedStringApp.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 9E8A7683B1C7BA67F4482933 /* libPods-LocalizedStringApp.a */; };
+		76F4720C6B697E9677FA72D7 /* libPods-LocalizedStringAppTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = B0E837DDBD8F520F45BD3C01 /* libPods-LocalizedStringAppTests.a */; };
+		E264FBA12319055A008E0DB5 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = E264FBA02319055A008E0DB5 /* AppDelegate.swift */; };
+		E264FBB82319055D008E0DB5 /* LocalizedStringAppTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E264FBB72319055D008E0DB5 /* LocalizedStringAppTests.swift */; };
+		E264FBCF2319082F008E0DB5 /* one.strings in Resources */ = {isa = PBXBuildFile; fileRef = E264FBCE2319082E008E0DB5 /* one.strings */; };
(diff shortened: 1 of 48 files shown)


tomlokhorst:
Note: As of iOS 13, end users can select the preferred language for an app in Settings.

abdelmagied94:
@tomlokhorst Could we set the preferred language in general instead of insert it in each call?

tomlokhorst:
No, that's not possible with this API.

abdelmagied94:
@tomlokhorst Thanks for your response! In my app, I let the user choose the language he wants in the start of the app. Therefore, I try to find a way to make this the default localization language. I see you set the `applicationLocale` with the first preferred language. Couldn't we set it dynamic?

IharAndreyeu:
Would be great if preferredLanguages: would be a method of string resource. I too face a situation where I need to change locale throughout the app, an the current API does not allow me to do that without every piece of the app knowing what the current locale is.

shadyk:
@IharAndreyeu @abdelmagied94  were you able to have work around regarding this?

theScud:
@tomlokhorst wanted to check with regards to the above comment, setting preferred language globally for the entire app. Is there anything in the roadmap to address this use case ?

mobilase:
Yesterday I change all my strings(~1k) to R.string in my app. In my app I can change language of app.
But after change strings to R.string - change language in app not working.
I need global change of locale. How I can do it? Variant with preferredLanguages does not suit me.

mobilase:
With help of [Article: Strong Typed Resources Management with R.swift](https://medium.com/@dayton159/strong-typed-resources-management-with-r-swift-1e3c8093ea41) - I found my workaround.

I create module Languages (where enum Languages and extension R)

```
import Foundation
import Localize_Swift
import Rswift

enum Languages: String {
    case ru
    case en
    case lt
    case lv
    case et
}

extension Languages {
    var langcode: String {
        switch self {
        case .en:
            return "en"
        case .ru:
            return "ru-RU"
        case .lt:
            return "lt-LT"
        case .lv:
            return "lv-LV"
        case .et:
            return "et-EE"
        }
    }
    
    static private var preffered: Languages = Languages.en
    static private var langBundle: Bundle?
    
    static func getLang() -> String {
        return preffered.langcode
    }
    
    static func setLang(_ langCode: String) {
        preffered = Languages(rawValue: langCode) ?? Languages.en
        let url = Bundle(for: R.ClassR.self).url(forResource: preffered.langcode, withExtension: "lproj")
        langBundle = Bundle(url: url!)
    }
    
    static func getBundle() -> Bundle? {
        if (langBundle == nil) {
            let url = Bundle(for: R.ClassR.self).url(forResource: preffered.langcode, withExtension: "lproj")
            langBundle = Bundle(url: url!)
        }
        return langBundle
    }
    
    static func getName(_ code: String) -> String {
        let locale : NSLocale = NSLocale(localeIdentifier: code)
        return locale.displayName(forKey: NSLocale.Key.identifier, value: code) ?? code
    }
    
    static func getLocalName(_ code: String) -> String {
        return Localize.displayNameForLanguage(code)
    }
}

extension R {
    static var text = R.string.localizable.self
    
    fileprivate class ClassR {}
    
    static func local(_ string: StringResource) -> String {
        return NSLocalizedString("\(string.key)", bundle: Languages.getBundle()!, comment: "")
    }
}
``` 

And when I change language of app I run this:
```
Languages.setLang(langCode)
```

And strings...
Before R.swift:
```
"test".localized()
```
With standard R.swift:
```
R.string.localizable.test()
```
With my workaround:
```
R.local(R.text.test)
```

PR #554, opened 2019-09-12
ResourceApp: Update path to rswift executable

mbernson:
It seems that SPM puts its build products in the `.build` directory now, which breaks ResourceApp.

### DIFF
--- a/ResourceApp/ResourceApp.xcodeproj/project.pbxproj
+++ b/ResourceApp/ResourceApp.xcodeproj/project.pbxproj
@@ -982,7 +982,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


tomlokhorst:
For developing R.swift, generate the `rswift.xcodeproj` project using:

```bash
swift package generate-xcodeproj --xcconfig-overrides RswiftConfig.xcconfig
````

As per https://github.com/mac-cain13/R.swift#building-from-source

This will configure `BUILT_PRODUCTS_DIR` to the `build` directory.

mbernson:
That commands takes care of `rswift.xcodeproj`, but doesn't seem to fix `ResourceApp.xcworkspace`, which isn't building for me.

tomlokhorst:
If you build rswift using the generated Xcode project, ResourceApp should work.

mbernson:
Aha, now it works. ✨ Thanks!

PR #555, opened 2019-09-12
SegueGenerator: Skip segues that have an empty identifier

### LINKED ISSUE #525: Warnings are generated for segues with an empty identifier where they shouldn't be
Steps to reproduce:

1. Create a 'show' segue or similar between two viewcontrollers in a storyboard
2. Give it an identifier
3. Clear the identifier field again

Expected result:

1. I don't get any warnings

Result:

1. Because the segue identifier in the storyboard XML is now an empty string instead of a null, I get the warning `[R.swift] Skipping 1 segue for 'FooController' because no swift identifier can be generated for segue: ''`

mbernson:
This PR causes the generator to skip any segues with an empty string identifier.

For some types of segues, such as from a UIButton to another view controller, a segue identifier is not required. When it **is** required, for example in a view controller initiated segue, Xcode will warn the user about it already, so R.swift doesn't need to warn about either case.

I'm not sure if I fixed it in the correct place, please let me know if you know a better way. :)

Closes #525


tomlokhorst:
Thanks for this PR!

Issue #563, opened 2019-10-02
Add Option to show development language defaults instead of keys if a translation is missing in a table.


scaio:
The current fallback behavior for `NSLocalizedString` checks if a table exists for a given locale but not if the key itself exists in the table. 

This means that if a key is missing (as often is the case for large products where the translations take a while to catch up) then the key itself will be returned unless a default value is specified, which currently only gets set for Base localization.

Projects using R.swift are prone to using keybased localization approaches, meaning keys are not actual English sentences but string ids (eg: `screenElement.someSubElement.mainText="Some very long winded text about something etc etc etc...";`). 

If you use key based localization and you have missing keys this effectively blocks you from releasing an app update until you get the translations, since you don't want users to see meaningless strings which is not ideal.

Arguably, if a release cannot wait for translations, the next best thing would be to at least attempt to default the missing keys to the development language so that user impact is minimized.

mac-cain13:
We're aiming to stay close to the behavior of Apples API. Is R.swift currently behaving different than using a plain call to `NSLocalizedString` would? I think `NSLocalizedString` would also return the key right and not the development language or am I mistaken?

scaio:
@mac-cain13 Thanks for replying :)

`NSLocalizedString` will return the key unless you pass a default value in case the key is not found. This allows one you use a non natural language key based approach with `NSLocalizedString` such as having the key be `my.buttton.text` (This seems to be the recommended approach by the r.swift docs). 

In an ideal world we should be able to halt releases until all translations are available. However,  that is often not feasible for production apps where you may need to ship a security update urgently or where you may want to run a beta tests before all translations are available. (especially if you have a broad range of supported languages!)

On an app using `NSLocalizedString` I would be able to use the `defaultValue` field to pass a natural language value in the development language to avoid having "garbage text" be visible to the user.

This PR aims to introduce this behavior **_as an opt in_** for users using non natural language keys, thus giving back some of the `defaultValue` functionality provided by `NSLocalizedString` that is currently obfuscated by `R.swift`.

Do you think there would be value in having this functionality in `R.swift`?

Is there another approach that you would recommend to support the use of non-natural language keys?

hujunfeng:
I agree with @scaio, and I would also like to see this option to be added. 

[...]

PR #564, opened 2019-10-02
Introduce useDevelopmentLanguageDefaults option

scaio:
## The problem this PR is trying to solve:
https://github.com/mac-cain13/R.swift/issues/563

## What this PR does:
This PR introduces a new flag `use-development-language-defaults` which when passed will cause the NSLocalizedString macros to have the `value:` option set to the translation value in the developmentLanguage if such a value is available and we are not using base localization.

## Testing
- I could really use some insight on how to properly unit test this behavior.

## End notes
First time opening a PR, please let me know if I missed something. I love this project and use it in pretty much all my apps. I think it would be great to have this functionality but please let me know if you think otherwise :)

### DIFF
--- a/Sources/RswiftCore/CallInformation.swift
+++ b/Sources/RswiftCore/CallInformation.swift
@@ -18,6 +18,7 @@ public struct CallInformation {
   let generators: [Generator]
   let accessLevel: AccessLevel
   let imports: [Module]
+  let useDevelopmentLanguageDefaults: Bool
 
   let xcodeprojURL: URL
   let targetName: String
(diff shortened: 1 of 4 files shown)


dphov:
@mac-cain13 Do you planned to include this feature to project?

snowman647:
@mac-cain13 would be nice to have this feature.

nikolayev:
It would be great if this makes into new release! Please take a look 🙏

mac-cain13:
If this makes us work more like Apples own API I would include this, but I think we're in line with `NSLocalizedString` now. We can discuss this further in https://github.com/mac-cain13/R.swift/issues/563

PR #579, opened 2019-11-04
Fix manual install example image

Eeyore741:
Fixed example image for manual install now has $SRCROOT var utilized (not $PODS_ROOT).


mac-cain13:
Thanks for the fix!

PR #588, opened 2019-11-30
Readme: suggest SPM as default install method

apptekstudios:
This PR adds install instructions for using R.Swift with SPM. This is preferable for those starting new projects and wanting to avoid cocoapods given the now-widespread support for SPM.

I changed it to be the recommended method, however opinions may differ... Please feel free to just add it as another option if that's preferable 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,11 +85,34 @@ Runtime validation with [`R.validate()`](Documentation/Examples.md#runtime-valid
 
 ## Installation
 
-[CocoaPods](http://cocoapods.org) is the recommended way of installation, as this avoids including any binary files into your project.
+**SwiftPackageManager** is the recommended way of installation, as this uses the built in XCode package manager.
 
-_Note on Carthage: R.swift is a tool used in a build step, it is not a dynamic library. Therefore [it is not possible](https://github.com/mac-cain13/R.swift/issues/42) to install it with Carthage._
+### SwiftPackageManager (recommended)
+#### 1) Add R.swift as a xcodeproject dependency
+1. In Xcode, go to File -> Swift Packages -> Add Package Dependency
+2. Paste the url of this repository: [https://github.com/mac-cain13/R.swift](https://github.com/mac-cain13/R.swift)
+3. Select 'Version: up to next major' and press *Next*
+4. Ensure that your targets are **not** ticked, and press **Finish**
 
-### CocoaPods (recommended)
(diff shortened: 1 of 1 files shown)


tomlokhorst:
Thanks for this PR!

I haven't tried this yet, but reading the PR, it seems interesting to use SPM in this way.
It does seem like quite a few steps to get everything working, but it's nice to not depend on CocoaPods anymore.

I'll have a more detailed look later on.

apptekstudios:
I opted to keep the instructions very verbose assuming it’s all new to people, but it boils down to very quick steps:
1) Create a Package.swift file
2) Make Run Script
3) Add R.swift.Library and generated file to target

mac-cain13:
Interesting to use SPM in this way, think we need to write it in a bit more compact way and not make it the default yet.

Also when we add this option think the readme needs a bit of restructuring or foldable parts to keep it readable.

Lets fix those points then merge this.

quentinfasquel:
@mac-cain13 I am not sure where to ask this.. important R.swift with SPM is fine, some people wrote about how to then use it in your Xcodeproj. However, I don't seem to find anyone who's written of using R.swift "outside an Xcode project". I am developing a swift package "Kit" that internally uses R.swift but I don't have a way to run R.swift for its sources without generating the Xcode project and adding a script phase. Is there any way I could do so without generating the Xcode project ?

fthdgn:
I don't like the BuildTools step. It is annoying to create folders on root project. This build script may be better.

```sh
cd $TEMP_DIR
mkdir RSwiftBinary
cd RSwiftBinary
echo `pwd`
echo "// swift-tools-version:5.1

import PackageDescription
let package = Package(
    name: \"RSwiftBinary\",
    platforms: [.macOS(.v10_11)],
    dependencies: [
        .package(url: \"https://github.com/mac-cain13/R.swift\", from: \"5.4.0\"),
    ]
)
" > Package.swift
/usr/bin/xcrun --sdk macosx swift run -c release rswift generate "$SRCROOT/R.generated.swift"
```

Another option is that.
Content of the Package.swift can be put on a folder on the library project. And we can run the script below. This method allow us to use always use correct R.swift and R.swift.Library library together.

```
cd "${BUILD_DIR%Build/*}/SourcePackages/checkouts/R.swift.Library/BuildTools/"
/usr/bin/xcrun --sdk macosx swift run -c release rswift generate "$SRCROOT/R.generated.swift"
```

PR #599, opened 2020-01-24
fix: url(with:) is only availale in iOS 9.0

m-colombo:
Used RSwift with SwiftUI, had to do this to make the preview works. Although I'm not sure this is the proper way of addressing the problem. (Did same PR also on XCodeEdit see https://github.com/tomlokhorst/XcodeEdit/pull/41)

### DIFF
--- a/Sources/RswiftCore/RswiftCore.swift
+++ b/Sources/RswiftCore/RswiftCore.swift
@@ -33,6 +33,7 @@ public struct RswiftCore {
     self.callInformation = callInformation
   }
 
+  @available(iOS 9.0, *)
   public func run() throws {
     do {
       let xcodeproj = try Xcodeproj(url: callInformation.xcodeprojURL)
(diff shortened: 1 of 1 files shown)


mac-cain13:
I'm not sure why this fix is necessary, since this is a change in the generator that only runs on macOS and you now add an available-annotation that is for iOS. But the generator never runs on iOS.

Are you sure you've integrated R.swift correctly into your project according to the Readme?

m-colombo:
I honestly don't know whether is my configuration wrong or it's the swiftUI preview tool.
A part from using swift package manager for both project (r.swift and r.swift.library) I followed the readme instruction, it worked both in emulator and physical device, but in the preview tool crashes.

mac-cain13:
Note that you should *NOT* add R.swift itself to your project using SPM, *only* the R.swift.library should be added as a dependency to your project.

m-colombo:
Ou ok, I saw the package.swift and I thought it was fine. Indeed importing it as a pod works. Thanks, sorry to mislead :)

PR #609, opened 2020-02-17
Only parse info and entitlements when generators are enabled

roelspruit:
This is a small change that make sure info plists and entitlement files are only parsed if the generators for those types are actually enabled. This is a workaround for an issue where URLs to entitlement files cannot be resolved at the moment.

Merging this PR would give users of the library the option to disable the entitlements generator if they don't need it and that way remove the warning  "[R.swift] File could not be parsed as InfoPlist from URL:  {entitlements url}"


tomlokhorst:
Great, thanks!

PR #611, opened 2020-02-26
Alternative for SPM dynamic linking

alexruperez:
We need [dynamic linking](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescription.md#methods-2) for this dependence.


alexruperez:
Sorry, my fault, not needed here, moved to https://github.com/mac-cain13/R.swift.Library/pull/38

PR #637, opened 2020-04-24
Support for namespaced file directories

apptekstudios:
This PR adds support for file resources that are in a Namespaced directory (ie. imported into the xcode project as a directory vs as a group).

For example, in one of my projects I include a directory of sound files that is included in my target. This addition means that R.Swift correctly recognises those files and provides safe access 👍


mac-cain13:
We're adding folder reference support for Xcode 16 now, that should fix files included in a reference instead of a group.

PR #639, opened 2020-04-25
Document SPM-only installation steps

tomlokhorst:
TODO
- [ ] Really cleanup the text
- [ ] Fix the wrongly indented steps
- [ ] Maybe reformat with some collapsable regions


AdieOlami:
Why do you have `PODS_ROOT` when you are not using cocoapods

Issue #559, opened 2019-09-19
R.color in watchOS (5.1.0.alpha.4)


xmanu:
Hello,

let me start with saying a sincere thank you for bringing watchOS support to R.swift. R.swift is one of my absolute go to libraries I always use and I was sad to resort to untyped code for watchOS until now.

I don't know if it is approriate to report "bug" for alpha versions, but I found an issue with watchOS support in 5.1.0.alpha.4.

When trying to use the usual syntax to get a UIColor `R.color.background()` this won't work on an watchOS target. The static let is correctly generated and the ColorResource is accessible, but the necessary static func to get the `UIColor` is still marked with `#if os(iOS) || os(tvOS)`.

Attached find a snippet of the generated code.

```
    #if os(iOS) || os(tvOS)
    /// `UIColor(named: "background", bundle: ..., traitCollection: ...)`
    @available(tvOS 11.0, *)
    @available(iOS 11.0, *)
    static func background(compatibleWith traitCollection: UIKit.UITraitCollection? = nil) -> UIKit.UIColor? {
      return UIKit.UIColor(resource: R.color.background, compatibleWith: traitCollection)
    }
    #endif
```

Thanks in advance for the great work. This saves me so much work on specific watchOS code I had to write before, especially because I can now use `R.string.`!

tomlokhorst:
R.color will be available as part of version 5.4.0

PR #642, opened 2020-04-28
Added color support for WatchOS

JeanAzzopardi:
I've run into the same issue as #559 where we're unable to use `R.colors.primary()` in our swift projects, as the function is not available for WatchOS targets.

I've modified the color generators to output colors for WatchOS as follows:

```
#if os(watchOS)
    /// `UIColor(named: "primary", bundle: ..., traitCollection: ...)`
    @available(watchOSApplicationExtension 4.0, *)
    static func primary(compatibleWith traitCollection: Any? = nil) -> UIKit.UIColor? {
      return UIKit.UIColor(named: R.color.primary.name)
    }
#endif
```

I'm using `traitCollection: Any?` parameter as I need to pass a parameter, otherwise Xcode will take the function to be a redefinition of the resource, in this case `R.colors.primary`. I had to pass in Any, as UITraitCollection is not available in watchOS.

I also extracted the color functions into `fileprivate` functions as they were being used twice in the same file.

### DIFF
--- a/Sources/RswiftCore/Generators/ColorStructGenerator.swift
+++ b/Sources/RswiftCore/Generators/ColorStructGenerator.swift
@@ -9,6 +9,58 @@
 
 import Foundation
 
+fileprivate func colorFunction(for name: String, at externalAccessLevel: AccessLevel, prefix: SwiftIdentifier) -> Function {
+  let structName = SwiftIdentifier(name: name)
+  let qualifiedName = prefix + structName
+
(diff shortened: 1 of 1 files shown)


tomlokhorst:
Thanks for this PR!

This is a nice fix, I've changed the unneeded parameter to be `(_: Void = ())`, similar to how the rest of R.swift works.

This will be released as part of version 5.4.0.

PR #644, opened 2020-04-29
Implement notarizing

tomasharkema:
Notarize binary with Apple.

As of alpha.6, release succeeds!
https://github.com/mac-cain13/R.swift/runs/630187426

https://github.com/mac-cain13/R.swift/releases/tag/v5.2.2-alpha.6 contains signed and notarized pkg and zip!


.github/workflows/release.yml
```diff
@@ -22,23 +22,85 @@ jobs:
...
+      uses: apple-actions/import-codesign-certs@v1
+      with: 
+        p12-file-base64: ${{ secrets.CERTIFICATES_APPLICATION_P12 }}
+        p12-password: ${{ secrets.CERTIFICATES_APPLICATION_P12_PASSWORD }}
+        create-keychain: true
+        keychain-password: ${{ secrets.APPLE_TEAM_ID }}
+    - name: Import Installer Siging
+      uses: apple-actions/import-codesign-certs@v1
+      with: 
+        p12-file-base64: ${{ secrets.CERTIFICATES_INSTALLER_P12 }}
```
mac-cain13, line 35: Think it's cleaner to put all certs in one file so we just need one action.

.github/workflows/release.yml
```diff
@@ -22,23 +22,85 @@ jobs:
...
+        create-keychain: false
+        keychain-password: ${{ secrets.APPLE_TEAM_ID }}
+    - name: Code Sign
+      run: |
+        codesign --force --options runtime --sign 'Developer ID Application: Mathijs Kadijk (5Z49PA849J)' .build/release/rswift
     - name: Archive
-      run: zip --junk-paths ${{ runner.temp }}/archive.zip .build/release/rswift License
-    - name: Upload
+      run: zip --junk-paths ${{ runner.temp }}/rswift-${{ github.event.release.tag_name }}.zip .build/release/rswift License
+    - name: Package
```
mac-cain13, line 44: Does providing a PKG have advantages? We didn't until now, so I feel maybe we should just keep it at a zip only?

.github/workflows/release.yml
```diff
@@ -22,23 +22,85 @@ jobs:
...
       uses: actions/upload-artifact@v1
       with:
         name: rswift-${{ github.event.release.tag_name }}
         path: .build/release/rswift
+    - name: Store artifact PKG
+      uses: actions/upload-artifact@v1
+      with:
+        name: rswift-${{ github.event.release.tag_name }}
+        path: ${{ runner.temp }}/rswift-${{ github.event.release.tag_name }}.pkg    
+    - name: Store artifact ZIP
```
mac-cain13, line 99: Maybe only the notarized zip is good enough as build artifact?

PR #673, opened 2020-10-15
Adds opt-in environment variable to change hosting bundle name

maciejpiotrowski89:
# What
This pull request adds an opt-in `HOSTING_BUNDLE_NAME` environment variable. The value set with the variable will be used as a bundle name used in the `R.generated.swift` file. A fallback to a standard value is maintained. If no value is passed a standard value is used.

This should be a non-breaking change.

The change is inspired by the rejected PR https://github.com/mac-cain13/R.swift/pull/395

# Why
This change is introduce to support the following scenerio:

An Xcode workspace consists of multiple projects. The main project contains an application target which depends on other projects. The projects contain static frameworks with resources. The resources for each framework are wrapped in `.bundle` wrapper and are embedded in the framework.

All frameworks are linked statically to the application and their bundles are copied to the application wrapper (`.app`). Source code of the frameworks uses R.swift library to access its resources. In order to generate proper access to the resources a bundle with proper name must be used. With the change in this Pull Request the name of a bundle is passed with `HOSTING_BUNDLE_NAME` environment variable.


tomlokhorst:
Thanks for this PR.

I'm a bit unsure about the project setup in which this is required.
Can you add documentation and an example project that uses this?

paciej00:
Sure, I will add them. Thanks for having a look on the PR!

tomlokhorst:
Thanks!
And be sure to merge in the latest master branch, I've refactored the example projects to be in a different directory.

paciej00:
@tomlokhorst I added a sample project depicting how the suggested change to the tool is used. Sorry for the delay.

sree127:
Heya @maciejpiotrowski89, I'm exactly in the same position as you are :) I was wondering when would this PR be merged? @tomlokhorst

maciejpiotrowski89:
@tomlokhorst could you approve running a workflow for this PR and have a look if the proposed changes can be merged into the main branch?

tomlokhorst:
Thanks again for your PR and the provided examples, this makes it clear.

I've update the code slightly to make it more consistent with the rest. The R.swift options are available as command line options. Only the variables that are provided by Xcode are read from the environment. 

If you're ok with these changes, I'll release this as version 6.1.

maciejpiotrowski89:
Your changes look fine to me @tomlokhorst , thanks for updating the code and examples.

Issue #685, opened 2020-12-21
Add support for custom SFSymbols


lennartkerkvliet:
With iOS 13 came the introduction of SFSymbols. These symbols can be called with UIImage systemName. Another feature introduced was the functionality to create [your own custom symbols](https://developer.apple.com/documentation/uikit/uiimage/creating_custom_symbol_images_for_your_app) which can be added to the asset catalog. These icon are accessible using the normal UIImage named API.

However, it seems like these custom symbols are not supported and can't be called using the R.swift syntax. Is there any way to make this work?

ejensen:
PR #689 adds support for custom SFSymbols

lennartkerkvliet:
Super cool @ejensen, thank you so much

PR #689, opened 2021-02-17
Support custom symbol image sets

ejensen:
This patch adds support for custom symbol images [Apple Documentation - Creating Custom Symbol Images for Your App](https://developer.apple.com/documentation/uikit/uiimage/creating_custom_symbol_images_for_your_app)

Closes issue #685


tomlokhorst:
Thank you for this PR!

tylerjames:
What's the usage of these like? Are we able to access all the variants for the symbol set?

PR #705, opened 2021-05-28
Bundle resources support

ivanzoid:
Hi!

Please review/check this implementation of bundle resources support.

Only following types of resources are enabled for now: `image`, `string`, `color`, `file`, I've tested them as I could.

`font` support won't work, I think we need to use the following technique to register fonts: https://stackoverflow.com/a/40749435/299063

`nib`, `segue` and `storyboard` probably will work, but I didn't test it as I don't have a project using them near me now. If someone could test it, it would be nice.

`reuseIdentifier` - probably will work, not sure, needs testing.

`entitlements`, `info` — I guess we don't need these for bundles

`id` - not sure, needs testing.

_Note_: to test not enabled features, you need to enable them in `RswiftCore.swift:run()` (search for `supportedGenerators`)

_Note 2_: because bundle may fail to load at runtime, updated R.swift.Library is needed (where all Bundle references are changed to optional), here is corresponding PR: https://github.com/mac-cain13/R.swift.Library/pull/42
Other way could be force unwrap Bundle instance (in `R.bundle.<foo>.bundle` property), then no changes are needed in R.swift.Library. But I guess it is not right way to do as it could lead to crash if bundle is missing in runtime.

_Note 3_: I noticed that for color generation on watchOS an API without bundle support is used (`UIColor(named:)`), however I see that `UIColor(named:in:compatibleWith:)` is supported on watchOS 4.0+. I guess there are issues with it about which I am not aware. So I removed bundle color generation for watchOS for now.

### DIFF
--- a/Sources/RswiftCore/Generators/ColorStructGenerator.swift
+++ b/Sources/RswiftCore/Generators/ColorStructGenerator.swift
@@ -57,8 +57,13 @@ struct ColorStructGenerator: ExternalOnlyStructGenerator {
       implements: [],
       typealiasses: [],
       properties: colorLets,
-      functions: groupedColors.uniques.map { [ generateColorFunction(for: $0, at: externalAccessLevel, prefix: qualifiedName),
-                                               generateWatchOSColorFunction(for: $0, at: externalAccessLevel, prefix: qualifiedName)] }.flatMap { $0 },
+      functions: groupedColors.uniques.map { name -> [Function] in
+        var functions = [ generateColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName)]
+        if case .hostingBundle = bundle {
+          functions.append(generateWatchOSColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName))
+        }
+        return functions
+      }.flatMap { $0 },
       structs: structs,
       classes: [],
       os: []
@@ -101,7 +106,7 @@ private extension NamespacedAssetSubfolder {
           value: "Rswift.ColorResource(bundle: \(bundle), name: \"\(colorPath)\(name)\")"
         )
     }
-
+    
     return Struct(
       availables: [],
       comments: ["This `\(qualifiedName)` struct is generated, and contains static references to \(colorLets.count) colors."],
@@ -110,8 +115,13 @@ private extension NamespacedAssetSubfolder {
       implements: [],
       typealiasses: [],
       properties: colorLets,
-      functions: groupedFunctions.uniques.map { [ generateColorFunction(for: $0, at: externalAccessLevel, prefix: qualifiedName),
-                                                  generateWatchOSColorFunction(for: $0, at: externalAccessLevel, prefix: qualifiedName)] }.flatMap { $0 },
+      functions: groupedFunctions.uniques.map { name -> [Function] in
+        var functions = [ generateColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName)]
+        if case .hostingBundle = bundle {
+          functions.append(generateWatchOSColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName))
+        }
+        return functions
+      }.flatMap { $0 },
       structs: structs,
       classes: [],
       os: []
--- a/Sources/RswiftCore/Generators/ColorStructGenerator.swift
+++ b/Sources/RswiftCore/Generators/ColorStructGenerator.swift
@@ -16,7 +16,7 @@ struct ColorStructGenerator: ExternalOnlyStructGenerator {
     self.assetFolders = assetFolders
   }
 
-  func generatedStruct(at externalAccessLevel: AccessLevel, prefix: SwiftIdentifier, bundle: String) -> Struct {
+  func generatedStruct(at externalAccessLevel: AccessLevel, prefix: SwiftIdentifier, bundle: BundleExpression) -> Struct {
     let structName: SwiftIdentifier = "color"
     let qualifiedName = prefix + structName
     let assetFolderColorNames = assetFolders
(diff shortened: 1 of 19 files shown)


Sources/RswiftCore/Generators/StringsStructGenerator.swift
```diff
@@ -285,6 +308,43 @@ struct StringsStructGenerator: ExternalOnlyStructGenerator {
       defaultValue: "nil"
     )
     params.append(prefereredLanguages)
+    
+    let body: String
```
CaioSym, line 312: This does not seem very DRY. Is there a way for us to isolate just the part that changes? (ie: the nullability check?) Maybe consider composing body in two steps where in the first step you define
```swift
var body: String
switch bundle {
    case .hostingBundle:      
        body = """
        let bundle = \(bundle)
        """
    case .customBundle:
      body = """
        guard let bundle = \(bundle) else {
          return "\(values.key.escapedStringLiteral)"
        }
       """
    }

body += """
 
       guard let preferredLanguages = preferredLanguages else {
          let format = \(values.swiftCode(bundle: "bundle"))
          return String(format: format, locale: applicationLocale, \(args))
        }
        guard let (locale, foundBundle) = localeBundle(parentBundle: bundle, tableName: "\(values.tableName)", preferredLanguages: preferredLanguages) else {
          return "\(values.key.escapedStringLiteral)" 
        }
        let format = \(values.swiftCode(bundle: "foundBundle"))
        return String(format: format, locale: locale, \(args))
        """
    }
```

You could even move the switch into an extension of BundleExpression to avoid having to type it twice then!

CaioSym (commented):
Thanks for putting this together. I would love to see something like this make it into the library. Added some comments to help out with reviewing.

Sources/RswiftCore/Util/Struct+InternalProperties.swift
```diff
@@ -64,6 +64,7 @@ extension Struct {
         name: "localeBundle",
         generics: nil,
         parameters: [
+          .init(name: "parentBundle", type: Type._Bundle),
```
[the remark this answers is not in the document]
ivanzoid: When there are multiple variables of the same type in the same namespace, I prefer to give all of them more detailed names to prevent possible wrong usage in the future when code is changed/refactored.

ivanzoid:
@CaioSym Added fixes!

Sources/RswiftCore/Generators/ColorStructGenerator.swift
```diff
@@ -57,8 +57,13 @@ struct ColorStructGenerator: ExternalOnlyStructGenerator {
       implements: [],
       typealiasses: [],
       properties: colorLets,
-      functions: groupedColors.uniques.map { [ generateColorFunction(for: $0, at: externalAccessLevel, prefix: qualifiedName),
-                                               generateWatchOSColorFunction(for: $0, at: externalAccessLevel, prefix: qualifiedName)] }.flatMap { $0 },
+      functions: groupedColors.uniques.map { name -> [Function] in
```
yegorsch, line 60: Curious if we can `flatMap` directly here like this:
```suggestion
      functions: groupedColors.uniques.flatMap { name -> [Function] in
```

Sources/RswiftCore/Generators/ColorStructGenerator.swift
```diff
@@ -57,8 +57,13 @@ struct ColorStructGenerator: ExternalOnlyStructGenerator {
...
+        var functions = [ generateColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName)]
+        if case .hostingBundle = bundle {
+          functions.append(generateWatchOSColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName))
+        }
+        return functions
+      }.flatMap { $0 },
```
yegorsch, line 66: Curious if we can flatMap directly here like this:
```suggestion
      functions: groupedColors.uniques.flatMap { name -> [Function] in
        var functions = [ generateColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName)]
        if case .hostingBundle = bundle {
          functions.append(generateWatchOSColorFunction(for: name, at: externalAccessLevel, prefix: qualifiedName))
        }
        return functions
      },
```

Sources/RswiftCore/ResourceTypes/Resources.swift
```diff
@@ -52,7 +54,14 @@ struct Resources {
 
       // All previous assets can also possibly be used as files
       if let resourceFile = tryResourceParsing({ try ResourceFile(url: url) }) {
-        resourceFiles.append(resourceFile)
+        if resourceFile.pathExtension == "bundle" {
+          let bundle = Bundle(bundleUrl: url, fileManager: fileManager)
+          if bundle.resources.resourceFiles.count != 0 {
```
yegorsch, line 59:
```suggestion
          if !bundle.resources.resourceFiles.isEmpty {
```

Sources/RswiftCore/ResourceTypes/Bundle.swift
```diff
@@ -0,0 +1,88 @@
...
+fileprivate extension URL {
+  func isDirectory() -> Bool {
+    do {
+      let value = try resourceValues(forKeys:[.isDirectoryKey])
+      return value.isDirectory ?? false
+    }
+    catch {
+      return false
+    }
+  }
```
yegorsch, line 87: If we are not handling an error here, can we just: 
```suggestion
  func isDirectory() -> Bool {
      let value = try? resourceValues(forKeys:[.isDirectoryKey])
      return value.isDirectory ?? false
  }
```

Sources/RswiftCore/ResourceTypes/Bundle.swift
```diff
@@ -0,0 +1,88 @@
...
+    let resourceDirectoryTypes: [WhiteListedExtensionsResourceType.Type] = [AssetFolder.self]
+    var resourceDirectorySuffixes = Set<String>()
+    resourceDirectoryTypes.forEach { resourceType in
+      resourceType.supportedExtensions.forEach { ext in
+        resourceDirectorySuffixes.insert(ext)
+      }
+    }
+
+    var prefixesToSkip = Set<String>()
+    prefixesToSkip.insert(".") // Ignore files like .DS_Store, etc.
```
yegorsch, line 38: nit: I'd move it inside of `shouldSkip` closure if it's not used anywhere else

rakuyoMo:
What is the function that this PR can achieve, please? Can you briefly describe it?

PR #767, opened 2022-11-14
R.generated file do not need to import both Foundation and UIKit

sookim-1:
- If **R.generated** file needs Foundation types, import Foundation.
- If **R.generated** file needs UIKit types (they all start with `UI`), import UIKit.

UIKit itself imports Foundation

Therefore if a file imports UIKit, it does not need to import Foundation explicitly
uikit


tomlokhorst:
Why do you think this change is needed?
Does it change the way generated code is compiled, or is it an esthetic change?

sookim-1:
> Why do you think this change is needed? Does it change the way generated code is compiled, or is it an esthetic change?

Just as using Foundation does not import Swift Standard Library, UIKit includes Foundation, so I think it is right to import only UIKit without specifying Foundation and UIKit.

tomlokhorst:
Deprecated by 7.0.0 release

Issue #454, opened 2018-09-28
Support for Data Assets


417-72KI:
###  Introduction
**Asset Catalog** supports **Data Assets** from **macOS 10.11/iOS 9.0**.
**Data Assets** can use with `NSDataAsset`.

Here is a document.
https://developer.apple.com/documentation/uikit/nsdataasset

#### Now
I'm using `animation.gif` in `Assets.xcassets` with below.
```swift
if let asset = NSDataAsset(name: "animation.gif") {
    let data = asset.data
    let animationImage = imagesFromAnimationGif(data)
}

func imagesFromAnimationGif(_ data: Data) -> UIImage {
    // parse data
}
```
#### Hope
I want call `animation.gif` via R.swift like below.
```swift
let data = R.data.animation_gif
let animationImage = imagesFromAnimationGif(data)
```

giulio92:
I would also like this feature!

isaac-weisberg:
In real programming, usage of Data Assets is immensely rare, and right now, for the first time in 5 years I would've had use of this feature, probably for the last time in my life :)

PR #773, opened 2022-12-01
Add function to get `Data` from `DataResource` directly

417-72KI:
Addresses #454 

There is only a simple initializer of `NSDataAsset` currently.

```swift
if let asset = NSDataAsset(resource: R.data.animation) {
    let data = asset.data
}
```

This PR makes it possible to call like `let data = R.data.animation()`


tomlokhorst:
Thanks again!

PR #775, opened 2022-12-01
Bump actions

uhooi:


tomlokhorst:
Thanks!

tomlokhorst:
I've undone these changes, since they were causing issues. 
One of the issues is that some updated actions don't work on symlinks so the workflows didn't actually work anymore.

PR #782, opened 2022-12-07
Update Migration Guide

ykws:
Remove R.generated.swift on Xcode.

Because the following error.

Filename "R.generated.swift" used twice


tomlokhorst:
Thank you for the PR!

Yes, this was mentioned in the [demo video](https://www.youtube.com/watch?v=icihJ_hin3I), but not in the actual text of the migration guide.

PR #789, opened 2022-12-11
Update GitHub workflows

tomlokhorst:


tomlokhorst:
The check workflow is now warning free.

The release workflow still has some warnings that need to be resolved (in a next PR).

PR #809, opened 2023-02-15
Allow main let to not be generated by providing omit-main-let flag

jimmya:
The app I'm working in is translated dynamically. So I'm relying on `preferredLanguages` to ensure the correct translations are loaded for the selected language. 

However, since the main let (`let R = _R(bundle: Bundle(for: BundleFinder.self))`) is always generated and available for everyone to use, it's quite hard to police proper usage of the custom definitions.

This non breaking PR adds a flag to omit the generation of the main let. This allows a custom `let R` to be defined with the customisation required. And will prevent accidental usage of the main let without customisations.


jimmya:
@mac-cain13 @tomlokhorst sorry to bother you but do you perhaps have time to take a look at this? It's sort of blocking us from migrating to 7.x at the moment.

tomlokhorst:
Thanks for this PR! This looks like a good feature for exactly this use case.

Sorry for not replying sooner, I just realised al my R.swift related emails were redirected to a different folder in my mailbox that I didn't check 🤦‍♂️

I'll merge this and will release it in a few days.

PR #811, opened 2023-02-20
Compare URLs with resolved symlinks

cerisier:
This fixes a base path mismatch when assets URLs are located in `/private/tmp`.
See https://stackoverflow.com/questions/43031045/for-what-paths-is-resolvingsymlinksinpath-wrong


tomlokhorst:
Thanks for this PR!

PR #816, opened 2023-03-16
Fix example testcase in Examples.md

null-z:
Fix the example testcase so that it works and is not confusing.


tomlokhorst:
Thanks!

Issue #819, opened 2023-03-27
.rswiftignore does not work correctly with Swift Package


manicmaniac:
R.swift cannot find .rswiftignore for a project that isn't made by xcodeproj but Package.swift.

It is because that `SOURCE_ROOT` environment variable doesn't exist when a project is build with Package.swift and currently the source code references `./SOURCE_ROOT` literally.

https://github.com/mac-cain13/R.swift/blob/7.3.0/Sources/rswift/App.swift#L91

tomlokhorst:
Thanks for the PR, this will be released in the next version.

PR #820, opened 2023-03-27
Fix sourceRootURL

m-sugawara:
R.swift cannot find .rswiftignore for a project that isn't made by xcodeproj but Package.swift.

It is because that `SOURCE_ROOT` environment variable doesn't exist when a project is build with Package.swift and currently the source code references `./SOURCE_ROOT` literally.

issue: #819

### DIFF
--- a/Sources/rswift/App.swift
+++ b/Sources/rswift/App.swift
@@ -88,7 +88,7 @@ extension App {
             let sourceTreeURLs = SourceTreeURLs(
                 builtProductsDirURL: URL(fileURLWithPath: processInfo.environment[EnvironmentKeys.builtProductsDir] ?? EnvironmentKeys.builtProductsDir),
                 developerDirURL: URL(fileURLWithPath: processInfo.environment[EnvironmentKeys.developerDir] ?? EnvironmentKeys.developerDir),
-                sourceRootURL: URL(fileURLWithPath: processInfo.environment[EnvironmentKeys.sourceRoot] ?? EnvironmentKeys.sourceRoot),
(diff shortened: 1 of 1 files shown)


Sources/rswift/App.swift
```diff
@@ -88,7 +88,7 @@ extension App {
             let sourceTreeURLs = SourceTreeURLs(
                 builtProductsDirURL: URL(fileURLWithPath: processInfo.environment[EnvironmentKeys.builtProductsDir] ?? EnvironmentKeys.builtProductsDir),
                 developerDirURL: URL(fileURLWithPath: processInfo.environment[EnvironmentKeys.developerDir] ?? EnvironmentKeys.developerDir),
-                sourceRootURL: URL(fileURLWithPath: processInfo.environment[EnvironmentKeys.sourceRoot] ?? EnvironmentKeys.sourceRoot),
+                sourceRootURL: URL(fileURLWithPath: processInfo.environment[EnvironmentKeys.sourceRoot] ?? ""),
```
m-sugawara, line 91: https://developer.apple.com/documentation/foundation/url/3126800-init

> If the path is an empty string, the system interprets it as “.”.

tomlokhorst: Then why not write `?? "."` here? 
That way you don't need to link to the documentation to explain the code, and it does the same thing, right?

tomlokhorst:
Thanks! This seems like a good fix. Will be in the next release.

Sources/rswift/App.swift
m-sugawara: Indeed, that might have been a better option.
Thank you for your review.

Issue #812, opened 2023-03-02
Localised strings with arguments are improperly formatted due to missing locale


glaurent:
It seems in version 7.2.4, strings like "this is an integer %d" will no longer be formatted properly because String(format:) is passed a nil locale. The fix is that at this line

https://github.com/mac-cain13/R.swift/blob/62a6b011b2f8f986f3c4665d801739849e814bae/Sources/RswiftResources/Integrations/StringResource%2BIntegrations.swift#L17

you should have

`self = String(format: format, locale: overrideLocale ?? Locale.current, arguments: arguments)`

tomlokhorst:
I see that this is indeed the case.

I'll have to investigate this further to see if this is a new issue that started with the 7.0 refactor, or if this has always been the way R.swift behaves.

glaurent:
It's not there in 6.x (I discovered it when upgrading from 6.x to current version).

paul-lavoine:
Any news on that topic ?

nicolasgoutaland:
Any update on that @tomlokhorst ? Is it a confirmed issue ?

tomlokhorst:
This is fixed in 7.4.0

PR #839, opened 2023-07-24
Use current locale when overrideLocale is missing

ribilynn:
As described in issue #812, when localizing strings containing arguments like "My number: %d", Rswift does not use the locale of the current environment when the locale argument is missing (which is the default behavior). So the string cannot be formatted correctly.

For example, in `Localizable.strings` in Japanese, when using %d, a comma should be inserted every three digits:
```
"My number: %d" = "数値：%d";
```

Current Rswift gives us this:
```swift
R.string.localizable.myNumberD(1234567890)
// 数値：1234567890

R.string(locale: Locale(identifier: "ja")).localizable.myNumberD(1234567890)
// 数値：1,234,567,890
```

After PR:
```swift
R.string.localizable.myNumberD(1234567890)
// 数値：1,234,567,890

R.string(locale: Locale(identifier: "ja")).localizable.myNumberD(1234567890)
// 数値：1,234,567,890
```

As comparison, iOS's behaviour is
```swift
let num = 1234567890

Text("Text: \(num)")
// Text: 1,234,567,890

Text(String(format: "String format: %d", num))
// String format: 1234567890

Text(String.localizedStringWithFormat("localizedStringWithFormat: %d", num))
// localizedStringWithFormat: 1,234,567,890
```

### DIFF
--- a/Sources/RswiftResources/Integrations/StringResource+Integrations.swift
+++ b/Sources/RswiftResources/Integrations/StringResource+Integrations.swift
@@ -14,7 +14,7 @@ extension String {
         case let .hosting(bundle):
             // With fallback to developmentValue
             let format = NSLocalizedString(key.description, tableName: tableName, bundle: bundle, value: developmentValue ?? "", comment: "")
-            self = String(format: format, locale: overrideLocale, arguments: arguments)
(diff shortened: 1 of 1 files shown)


tomlokhorst:
Thank you for this PR!

PR #842, opened 2023-08-05
Support custom localized string loading strategies

liamnichols:
WIP Implementation for #843


tomlokhorst:
See: https://github.com/mac-cain13/R.swift/issues/843#issuecomment-1712923342

PR #854, opened 2023-09-28
Conflict with SwiftUI: Renaming Classes to RImageRessource and RColorRessource

Jewelch:
### Problem:

Apple has introduced two new classes, **ImageResource** and **ColorResource**, in SwiftUI. These classes have the same names as our existing classes, **ImageRessource** and **ColorRessource**, which has led to a naming conflict.

### Solution:
To avoid this conflict and ensure the smooth functioning of our package, I have renamed our classes as follows:

- _ImageRessource_ is now **RImageRessource**
- _ColorRessource_ is now **RColorRessource**

### Why This Change is Necessary:
By renaming our classes, we prevent any ambiguity when developers use our package alongside SwiftUI. This change aligns with best practices to avoid naming conflicts and ensures that our package remains compatible with the latest iOS SDK.

### Links to Apple Documentation:

[Apple's ImageResource Documentation](https://developer.apple.com/documentation/developertoolssupport/imageresource)
[Apple's ColorResource Documentation](https://developer.apple.com/documentation/developertoolssupport/colorresource)

### Impact:
This change may require updates in the codebase where our package is used. Developers will need to update references to the old class names (_ImageRessource_ and _ColorRessource_) to the new names (**RImageRessource** and RColorRessource) to ensure compatibility.

### DIFF
--- a/Sources/RswiftGenerators/AssetCatalog+Generator.swift
+++ b/Sources/RswiftGenerators/AssetCatalog+Generator.swift
@@ -13,7 +13,7 @@ public protocol AssetCatalogContent {
     func generateVarGetter() -> VarGetter
 }
 
-extension ColorResource {
+extension RColorResource {
     public static func generateStruct(catalogs: [AssetCatalog], prefix: SwiftIdentifier) -> Struct {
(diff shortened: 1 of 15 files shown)


mac-cain13:
Since these are both in separate frameworks and the new Xcode versions are better in communicating errors around this I think we can keep it as is for now.

PR #869, opened 2023-11-24
Register fonts on-demand

cyrilmottier:
The whole idea of this change is to make https://developer.apple.com/documentation/uikit/text_display_and_fonts/adding_a_custom_font_to_your_app completely obsolete. Having the font part of the “UIAppFonts” value in Info.plist breaks the whole purpose of R.Swift (runtime vs compile-time behaviour) and is very error-prone

R.Swift does provide a `R.validate()` runtime validation to counter this problem. This is unfortunately a runtime only approach.

This change simply register the font on-demand. Note that it continuously calls FontResource.canBeLoaded considering it as cheap (which is the case based on my experience) but we could keep some state somewhere if necessary.


cyrilmottier:
👋 Should I ping someone in particular ?

PR #885, opened 2024-02-28
[Tooling] Mise support

mustiikhalil:
Adds basic support for [mise](https://mise.jdx.dev) which allows versioning depending on directories


mac-cain13:
Thanks for the PR.

I'm sorry, but we're moving towards less package managers being explicitly supported. We want SPM to be the default that we focus on and Cocoapods the fallback for older projects.

PR #887, opened 2024-03-10
Add beautiful logo for readme on repo

aiKrice:
This repo needs to be more beautiful <3 .

Explanation: 
The sparrow which is the symbol of swift and the green color is the wink of Android 
Generated by AI


mac-cain13:
Thanks for generating the image! However I think we're fine without a logo for now, if we add one I want to put some thought into what it should look like first before adopting anything.

PR #890, opened 2024-03-24
`StoryboardViewControllerIdentifier.callAsFunction(creator:)` should return non-optional value

hujunfeng:
This method calls `UIStoryboard.instantiateViewController(identifier:creator:)` internally, which returns a non-optional view controller. Therefore, it is unnecessary to return optional value in the `StoryboardViewControllerIdentifier` method.

See also: https://developer.apple.com/documentation/uikit/uistoryboard/3213989-instantiateviewcontroller


tomlokhorst:
Thanks!

Issue #860, opened 2023-10-16
How can use `--generators` in swift plugin for SPM


nshapovalov:
https://github.com/mac-cain13/R.swift/blob/main/Documentation/Ignoring.md#only-run-specific-generators-exclude-rsomething

I'd like to add `--generators image,string` to my module, how can I do it with plugin in package.swift?

erolando:
+1

mac-cain13:
This is indeed not possible at the moment, good feature request though. :)

PR #896, opened 2024-06-20
Add support for `rswift.json` config to BuildToolPlugin

TheNightmanCodeth:
I noticed there was no way to set custom arguments to the generation command when using SPM or Xcode BuildToolPlugin

This just allows for configuration via a `rswift.json` file placed at the project root (or elsewhere if it is added manually to the Xcode project)

The config json would look like:

``` json
{
    "generators": ["image", "string"],
    "rswiftignorePath": "/Users/TheNightmanCodeth/project/other/.rswiftignore",
    "additionalArguments": ["--somethingelse", "value"]
}
```

or any variation (ie. only additionalArguments, only generators and rswiftignorePath, etc.)

Addresses #860 (among others)

### DIFF
--- a/Plugins/RswiftGenerateInternalResources/RswiftGenerateInternalResources.swift
+++ b/Plugins/RswiftGenerateInternalResources/RswiftGenerateInternalResources.swift
@@ -8,6 +8,19 @@
 import Foundation
 import PackagePlugin
 
+struct RSwiftConfig: Codable {
+    enum Generator: String, Codable {
+        case image, string, color
+        case file, font, nib
+        case segue, storyboard, reuseIdentifier
+        case entitlements, info, id
+    }
+    
+    let generators: [Generator]?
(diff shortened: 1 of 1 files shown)


fthdgn:
This config file is used only by plugin defined on RswiftGenerateInternalResources.swift. There is also plugin on RswiftGeneratePublicResources.swift.

These two plugins may be combined and access level is set by rswift.json. But for backward compatibility, it is better to not delete old plugins.

TheNightmanCodeth:
Yeah good call this was a quick solution for a narrow use-case. There's another MR #899 raised just after this that, at a quick glance, appears to fill in those gaps. 

I'm going to go ahead and close this since I'm no longer using R.swift and don't plan on coming back to finish this out. If someone else would like to take what I've done and finish it in another MR that's fine with me.

PR #899, opened 2024-07-11
Added the ability to specify options, or default arguments, for plugin invocations

SomeRandomiOSDev:
## What does this PR do?

This PR fills in a gap that is currently present when using the Package/Xcode Plugins to perform source code generation, either manually using the `RswiftGenerateResourcesCommand`, or automatically using the `RswiftGeneratePublicResources` or `RswiftGenerateInternalResources` plugins. The gap to which I speak is that if any customization is needed, i.e. `rswift` arguments, that can _only_ be done when manually running the `RswiftGenerateResourcesCommand` plugin and the arguments must be specified _each_ time the plugin is used.

This is obviously not an ideal scenario, so this PR seeks to alleviate this issue by adding support for an "options" file. The way in which this works is slightly different for Xcode Plugins vs. SPM Plugins, but the premise and formatting is the same:

1. A `.rswiftoptions` file is declared in your sources root (Xcode Plugins can create an additional file in the root of the project as well)
2. When either of the plugins mentioned above the options are loaded from this options file, validated, and then combined with any defaults and options provided via the plugin invocation and including those in the options provided to the `rswift` utility. 
3. That's it!

The `.rswiftoptions` file is a JSON file that has a very simple schema:

```json
{
  "generators": ["..."],
  "omit-main-let": true/false,
  "imports": ["..."],
  "access-level": "public"/"internal",
  "rswiftignore": "...",
  "bundle-source": "finder"/"module",
  "output-path": "...",
  "additional-arguments": ["..."]
}
```

All of the fields in this JSON object are optional and are applied in a systemic and deterministic way when executing the plugins:

1. Create an options structure any arguments passed in from the plugin invocation parsed out. Options provided to the plugin invocation _always_ take precedent since they have to be passed in for each execution, indicating that the intent is to override any preset or default values.
2. (For Xcode Plugins) Attempt to load a `.rswiftoptions` file that is located in the same directory as the project file invoking the plugin. These options are combined with the options in the previous step in such a way to where the previous options will not be overridden by the new options, should a particular option already be set.
3. Attempt to load a `.rswiftoptions` file that is located in the root directory containing the source files for the given target. These options are combined with the options in the previous step in such a way to where the previous options will not be overridden by the new options, should a particular option already be set.
4. Lastly, apply any "fallback" options, i.e. options that should be used should the options not be specified by any other means.

When done in this manner, this allows for one to use these options files to specify the arguments that should always be provided, but allows for those values to be overridden if invoking the plugin directly.

## How are these changes implemented?

The way in which the implementation is done is multifaceted:

1. First, a new library target was created called `RswiftShared` to contain code that should be shared between the `rswift` executable target as well as the plugin targets. The `AccessLevel`, `BundleSource`, and `ResourceType` types were all moved into this new target.
2. Next, symbolic links were created in the plugin sources directories to reference the `RswiftShared` sources directory. This is done to get around an issue where plugin targets cannot have a dependency on a library target so the sources are brought into each of the plugin targets via this symbolic link.
3. An additional type was created to mimic the structure of the JSON options file with a conformance to `Decodable` for easy parsing.
4. Lastly, the implementation of each plugin was updated to attempt to load these `.rswiftoptions` files and combine these options with any provided from the plugin invocation, if relevant.

## Any testing done?

Yes, testing was performed locally using an Xcode project and a Swift Package, each of which was tested by using various combinations of the aforementioned plugins, options files, and options provided when manually invoking the `RswiftGenerateResourcesCommand`. All results of this testing align with the deterministic manner in which those options should have been applied, as described above

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,8 +23,9 @@ let package = Package(
     ],
     targets: [
         .target(name: "RswiftResources"),
+        .target(name: "RswiftShared"),
         .target(name: "RswiftGenerators", dependencies: ["RswiftResources"]),
-        .target(name: "RswiftParsers", dependencies: ["RswiftResources", "XcodeEdit"]),
+        .target(name: "RswiftParsers", dependencies: ["RswiftResources", "RswiftShared", "XcodeEdit"]),
 
         .testTarget(name: "RswiftGeneratorsTests", dependencies: ["RswiftGenerators"]),
         .testTarget(name: "RswiftParsersTests", dependencies: ["RswiftParsers"]),
@@ -33,6 +34,7 @@ let package = Package(
         .executableTarget(name: "rswift", dependencies: [
             .target(name: "RswiftParsers"),
             .target(name: "RswiftGenerators"),
+            .target(name: "RswiftShared"),
             .product(name: "ArgumentParser", package: "swift-argument-parser"),
         ]),
 
--- a/Plugins/RswiftGenerateInternalResources/RswiftGenerateInternalResources.swift
+++ b/Plugins/RswiftGenerateInternalResources/RswiftGenerateInternalResources.swift
@@ -13,33 +13,56 @@ struct RswiftGenerateInternalResources: BuildToolPlugin {
     func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
         guard let target = target as? SourceModuleTarget else { return [] }
 
-        let outputDirectoryPath = context.pluginWorkDirectory
-            .appending(subpath: target.name)
+        let defaultOutputDirectoryPath = context.pluginWorkDirectory.appending(subpath: target.name)
+        let rswiftPath = defaultOutputDirectoryPath.appending(subpath: "R.generated.swift")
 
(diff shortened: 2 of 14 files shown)


SomeRandomiOSDev:
@tomlokhorst Could I get a review on this PR?

SomeRandomiOSDev:
@tomlokhorst Can this PR get reviewed? It would be really nice to have this feature available more broadly and I think it'd be useful for general consumers of this library

KillaCrow (approved):

tomlokhorst:
Thank you for this PR!
I haven't reviewed it yet, we will review it when we have free time available.

mmuchaidze-tbc:
Hello everyone, what about this feature? When will it be merged any ideas?

SomeRandomiOSDev:
Hey @tomlokhorst , just checking to see if there’s any update with these changes?

SomeRandomiOSDev:
Hey @tomlokhorst, just a friendly nudge to see if we can move this PR along

Issue #902, opened 2024-07-30
Let 'R' is not concurrency-safe because non-'Sendable' type '_R' may have shared mutable state


jeannustre:
Hi,

We're getting a build error in R.generated.swift line 10 : 

> Let 'R' is not concurrency-safe because non-'Sendable' type '_R' may have shared mutable state

This happens on Xcode 16 Beta 4 (Swift 6), using either R.swift 7.5.0 or 7.5.1-alpha.3.

Lumisilk:
Just in case, for now you can workaround it by add this:
```swift
extension _R: @unchecked Sendable {}
```

tomlokhorst:
Fixed in release 7.6.0

jaltreuter:
In case anyone else is running into this issue even after updating, I'll note that this is fixed for me (I'm using 7.7.0) but I had to delete the contents of each of my `R.generated.swift` files in order to force re-generation of these files.  Must be some kind of caching issue?

PR #906, opened 2024-09-04
Add conformance to Sendable

g-laures:
Fix the following issue: https://github.com/mac-cain13/R.swift/issues/902

### DIFF
--- a/Sources/rswift/RswiftCore.swift
+++ b/Sources/rswift/RswiftCore.swift
@@ -201,7 +201,7 @@ public struct RswiftCore {
             valueCodeString: validateLines
         )
 
-        var s = Struct(name: structName, additionalModuleReferences: [.rswiftResources]) {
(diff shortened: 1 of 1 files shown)


mac-cain13:
Thanks for the PR, we will solve this by updating the code generator to generate slightly different code that makes the Sendable being inferred by the compiler.

letko-dmitry:
Marking all the types as `Sendable`, guarded by `#if compiler`, remains a better solution as it enables wider usage of the types, rather than simply silencing the warnings.

mac-cain13:
R.swift 7.6 is now released which makes the R type sendable.

letko-dmitry:
But, as far as I understand, to use types in modules, it's still necessary to mark them as `Sendable` explicitly.

letko-dmitry:
And it applies to all types, including those from `RswiftLibrary`.

tomlokhorst:
You are right @letko-dmitry! Thank you for following up on this.

When public structs are generated, it is needed to explicitly conform to Sendable.

(For internal types, the compiler can infer Sendable)

PR #916, opened 2024-10-14
Dependency conflicts

titlesxyz:
We're running into some dependency conflicts because the URLs in the `Package.resolved` file do not have the standard `.git` suffix.

If another dependency requests `https://github.com/apple/swift-argument-parser.git` then Xcode cannot resolve the `swift-argument-parser` library for either dependency.

By adding the `.git` suffix the dependency issue is fixed. 

Version: Xcode 16.0
Swift: Swift: 5.9
Platform: iOS 16

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -3,7 +3,7 @@
     {
       "identity" : "swift-argument-parser",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-argument-parser",
+      "location" : "https://github.com/apple/swift-argument-parser.git",
       "state" : {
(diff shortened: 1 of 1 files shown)


tomlokhorst:
Hi, I don't think adding a .git extension to repositories is a common convention. 
It's not something I see in Apple's open source SPM projects on GitHub.

It would seem to me if we added this, other people would run into dependency conflicts because of the uncommon URLs.

tomlokhorst:
Yes, it may be an issue with SPM.

I don't know if there is a way to solve this for all users.

For example if you also depend on Swift OpenAPI Generator, you'd have the same issue: https://github.com/apple/swift-openapi-generator/blob/main/Package.swift#L58

Issue #918, opened 2024-10-23
Plist resources do not work when the path includes arrays


jhildensperger:
For instance, the following test will fail when added to the ResourceApp example.

```swift
  func testUIApplicationSceneManifest() {
    XCTAssertFalse(R.info.uiApplicationSceneManifest.uiApplicationSupportsMultipleScenes)
    
    XCTAssertEqual(R.info.uiApplicationSceneManifest.uiSceneConfigurations.uiWindowSceneSessionRoleApplication.defaultConfiguration.uiSceneConfigurationName, "Default Configuration")
    XCTAssertEqual(R.info.uiApplicationSceneManifest.uiSceneConfigurations.uiWindowSceneSessionRoleApplication.defaultConfiguration.uiSceneDelegateClassName, "ResourceApp.SceneDelegate")
  }
```

`testUIApplicationSceneManifest(): XCTAssertEqual failed: ("$(PRODUCT_MODULE_NAME).SceneDelegate") is not equal to ("ResourceApp.SceneDelegate")`

This issue is mostly hidden by the fact that the value is copied as the default in the R.generated.swift output, however, this does not work properly for variable expansions.

PR #919, opened 2024-10-23
Fix support for plist items that are in arrays

jhildensperger:
Resolves https://github.com/mac-cain13/R.swift/issues/918

### DIFF
--- a/Examples/ResourceApp/ResourceAppTests/InfoPlistTests.swift
+++ b/Examples/ResourceApp/ResourceAppTests/InfoPlistTests.swift
@@ -19,4 +19,34 @@ class InfoPlistTests: XCTestCase {
   func testVariable() {
     XCTAssertEqual(R.info.nsExtension.nsExtensionPrincipalClass, "ResourceApp.IntentHandler")
   }
+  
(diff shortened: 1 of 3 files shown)


tomlokhorst:
Hi, thanks for this PR.
A good observation that the current behaviour of R.swift is incorrect for variable expansion in arrays.

However this PR doesn't seem to change anything about the behaviour?

If I run the tests, it still fails on:
`XCTAssertEqual(R.info.uiApplicationSceneManifest.uiSceneConfigurations.uiWindowSceneSessionRoleApplication.defaultConfiguration.uiSceneDelegateClassName, "ResourceApp.SceneDelegate")
  }`

jhildensperger:
Hi Tom, Did you build the release version rswift before running the test? While developing, I had the run script phase pointing to my build products, but didn't include that in the PR.

tomlokhorst:
Thank you for that, I indeed didn't build the release before hand. I forgot the ResourceApp is pointing to the explicit path.

Good debugging!

And a good fix. I'll merge it, it will be in the next release.

PR #920, opened 2024-11-01
Use files on PBXFileSystemSynchronizedRootGroups directories

fthdgn:
R.swift cannot detect files, assets and localization files inside of PBXFileSystemSynchronizedRootGroup directories.

This commit fixes this problem, by iterating these directories on the filesystem.

I need to update XcodeEdit project to access fullPath of these directories.

This is my PR on XcodeEdit https://github.com/tomlokhorst/XcodeEdit/pull/48


tomlokhorst:
Thank you for this PR!

I've merged your PR on XcodeEdit. 
To scan the file system for synchronized folders, I've implemented https://github.com/mac-cain13/R.swift/pull/921. That also supports folders that are not at the root level, and implements membership exceptions.

PR #922, opened 2024-11-14
Add `https` protocol to SPM Installation

ExTBH:
Xcode's SPM cant resolve packages without a protocol, since i copied the URL directly from here and it was just silently failing it took me 4 hours and a system update to MacOS Sequoia to catch it............


tomlokhorst:
Thanks for this PR!

Good to have these kinds of onboarding improvements.

PR #923, opened 2024-11-15
Accept optional [preferredLanguages]

ReDetection:
Hi. I've been happily using R.swift for a long time now, so I missed the whole 6.x version and now I am migrating from the version 5 to the version 7. Everything seems to be really smooth (thanks!) except one little thing: previously `preferredLanguages` parameter did accept `nil`, so I can pass it down my call chain if I want to, and now I can no longer do that. Funny enough, it seems every part of R.swift still accepts `preferredLanguages: nil` just fine, except the very top autogenerated file. Here is the PR to fix that. 

One note: I could not locate where the tests of those are located, I can only see the example app, which doesn't really test anything, so I am kinda confused here.


ReDetection:
@tomlokhorst I would appreciate a review 🙏

tomlokhorst:
Thank you for this PR!
I haven't reviewed it yet, we will review it when we have free time available.

